PR #3, opened 2024-06-11
DAO refactoring

w4terbomb:
Made the first part to refactor the code removing the DAO.

There are still some things like DB to refactor. /msyql driver etc.

Note: Check very well before merging, it may contain errors. If you find anything to improve (there are many), let me know, I can correct it and add other things.


Wern-rm:
Terrible MR - why delete the DAO if it's a programming pattern? You just removed the abstraction level and that's it - without doing anything good...

Wern-rm:
https://www.baeldung.com/java-dao-pattern

Wern-rm:
With this refactoring, you removed the ability to make a Postgres implementation instead of Mysql without breaking MySQL variations.

w4terbomb:
It's not finished yet Wern, I'm just trying to help. This was said to Neon.

Wern-rm:
> It's not finished yet Wern, I'm just trying to help. This was said to Neon.

We need to discuss this change in more detail, as it can both help and harm

neon-dev:
> Terrible MR - why delete the DAO if it's a programming pattern? You just removed the abstraction level and that's it - without doing anything good...

Please keep calm and be respectful.

> We need to discuss this change in more detail

Apparently, yes. I also believe there to be some kind of misunderstanding about this PR, as we all have the same goal in mind.

neon-dev (approved):

neon-dev:
Thank you for your contribution, highly appreciated!

PR #7, opened 2024-07-22
fix for data

sky1683589933:
Most of the modifications to some 'data' have been validated, and there are still some that need to be validated.


neon-dev:
Thanks for all the fixes.
Could you please create separate PRs per topic, each with a short description what was fixed? Like one PR for all quest related fixes, one for spawn coords, and so on.

While skimming through the changes I noticed many files with inconsistent formatting. You can check your IDE for EditorConfig support as explained in the readme and then auto-format the files.
I also noticed new Java quest handlers have an `@author Hilgert` tag. If you are not the author, please specify from which project you have taken these handlers from.

sky1683589933:
The new Java quest for the @ author Hilgert tag comes from this project. Since the dialog ID steps are the same, only the NPC ID and map ID need to be changed, so it is just borrowed, haha!

neon-dev:
I see. In this case, the author tag should not be transferred.
If you want, you can put your name there, including the other new Java handlers you created.

Is it ok for you to move part of the changes into new PRs?

sky1683589933:
Okay, I will add comments and categorize PR for each modification
I am a probie, and this matter may take some time for me, haha!
Since English is not my native language, annotations may be difficult to understand, that's it!

neon-dev:
Thank you and no worries. Take your time.

PR #23, opened 2025-01-26
Fixed Summon Cyclone Servant

crlfrancis:
Summoned servant should attack 3 times instead of 1


neon-dev:
This PR is on hold until we have figured out how the client and PTS data is supposed to map to the `attack_count`, as there is no clear relation. This indicates that the official server may not be using an attack count at all, but a timer for example. The other `<summonhoming/>` should also be checked.

PR #26, opened 2025-01-27
Fixed 3058 quest cannot be completed properly

sky1683589933:

### DIFF
--- a/game-server/config/main/gameserver.properties
+++ b/game-server/config/main/gameserver.properties
@@ -3,8 +3,8 @@
 # GS config's:
 # ----------------------------
 
-# GS Country Code (cc)
-# 1=NA, 2=EU, 7=RU, 99=Region free (allow any client)
+# Server country code (the client checks it against its cc start parameter)
+# 1=NA, 2=EU, 7=RU, 99=Region free (allows any client, but client will limit character names to 10 characters)
 # Default: 99
 gameserver.country.code = 99
 
--- a/game-server/src/com/aionemu/gameserver/configs/main/GSConfig.java
+++ b/game-server/src/com/aionemu/gameserver/configs/main/GSConfig.java
@@ -8,10 +8,9 @@ import com.aionemu.commons.configuration.Property;
 public class GSConfig {
 
 	/**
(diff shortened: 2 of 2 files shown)


game-server/data/handlers/quest/theobomos/_3006TheShugoFugitive.java
```diff
@@ -61,7 +61,7 @@ public boolean onDialogEvent(QuestEnv env) {
 							}
 							return false;
 						}
-						case SETPRO1: {
+						case SELECT2_1_1: {
```
neon-dev: `SETPRO1` was correct. Generally, SETPRO* actions are the ones that advance the quest step and I verified that this is the case here too. Did you maybe edit the wrong  quest file?

sky1683589933: Maybe? It may be related to the translation difference of "3006quest. html" on the CHS client.

neon-dev: Could be a problem in your language pack, yes. I cannot test it as I don't have this language pack. With the EU client, everything works normally using `SETPRO1`. Also, some quest dialog would be missing if using `SELECT2_1_1` here, which is not correct either as it ends the conversation with the npc early.

neon-dev: What you can (and should) change here is
```java
case SETPRO1:
	return defaultCloseDialog(env, 0, 1);
```
This advances the quest step and closes the quest dialog instead of keeping the window open.

sky1683589933: It's too late. Tomorrow I have time to test with "EU 3006quest. html". It seems that the conversation was interrupted due to a translation issue. In the CHS client, "SELECT2_1_1" must be used to continue the conversation.

sky1683589933: Okay, let's close it first.

PR #35, opened 2025-01-28
Fixed mob kill zone for quest 1604

sky1683589933:


neon-dev:
Please check your DMs, so you don't run into the same problems again and again 😄

PR #38, opened 2025-01-28
Fixed quest 3058 cannot be completed properly

sky1683589933:

### DIFF
--- a/game-server/data/handlers/quest/theobomos/_3058StoneofMabolo.java
+++ b/game-server/data/handlers/quest/theobomos/_3058StoneofMabolo.java
@@ -71,10 +71,7 @@ public class _3058StoneofMabolo extends AbstractQuestHandler {
 						return sendQuestDialog(env, 1352);
 					return false;
 				case SETPRO1:
-					if (var == 0) {
 						return defaultCloseDialog(env, 0, 1);
-					}
-					return false;
 			}
 		} else if (targetId == 203701) {
 			switch (env.getDialogActionId()) {
@@ -83,10 +80,7 @@ public class _3058StoneofMabolo extends AbstractQuestHandler {
 						return sendQuestDialog(env, 1693);
 					return false;
 				case SETPRO2:
-					if (var == 1) {
 						return defaultCloseDialog(env, 1, 2, true, false);
-					}
-					return false;
 			}
 		}
 		return false;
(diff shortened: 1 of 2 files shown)


game-server/data/handlers/quest/theobomos/_3058StoneofMabolo.java
```diff
@@ -68,10 +72,7 @@ else if (env.getDialogActionId() == SELECT_QUEST_REWARD)
 					return false;
 				case SETPRO1:
 					if (var == 0) {
```
neon-dev: You don't need to check the `var` here anymore, since you're using `defaultCloseDialog(env, 0, 1)` which checks the quest var internally.
This simplifies the whole `SETPRO1` case to a single return statement.

game-server/data/handlers/quest/theobomos/_3058StoneofMabolo.java
```diff
@@ -83,15 +84,30 @@ else if (env.getDialogActionId() == SELECT_QUEST_REWARD)
 					return false;
 				case SETPRO2:
 					if (var == 1) {
```
neon-dev: You don't need to check the `var` here anymore, since you're using `defaultCloseDialog(env, 1, 2)` which checks the quest var internally.
This simplifies the whole `SETPRO2` case to a single return statement.

neon-dev (changes requested):

game-server/data/handlers/quest/theobomos/_3058StoneofMabolo.java
sky1683589933: Okay, I understand. I will pay attention to other modifications next time because this local branch has already been deleted.

neon-dev: You can checkout the branch locally again, as long as it is not deleted on the remote origin (GitHub in this case).
You can also edit files directly in the browser:
grafik

neon-dev (approved):
I have made some changes based on this video I found: https://www.youtube.com/watch?v=x3QQfsMq7LM&t=12s
The quest start item now uses the correct animation packet and the last quest step now switches to the reward state while talking to Zirius.
Please test my changes and let me know if you think this pull request can be merged.

sky1683589933:
> I have made some changes based on this video I found: https://www.youtube.com/watch?v=x3QQfsMq7LM&t=12s The quest start item now uses the correct animation packet and the last quest step now switches to the reward state while talking to Zirius. Please test my changes and let me know if you think this pull request can be merged.

I have tested that this is a very good change, everything is perfect!
Adding items to the kernel and automatically matching animations of using items through the client is amazing!

PR #40, opened 2025-01-28
Fixed mob kill zone for quest 1604

sky1683589933:
Fixed the position of killing monsters, as monsters in the original position would be killed by NPCs, resulting in mission failure.


sky1683589933:
After retesting, it was found that luring the mob to the original kill zone could successfully complete the quest as long as it was not too far away from the mob, without the need to modify the kill zone.

sky1683589933:
Confirm that the kill zone does not need to be modified, I will close this pull request.

PR #41, opened 2025-01-29
Possible fix for AutoGroupService and SM Dispell

jvbhidalgo:
There is a problem that when 2 groups of the same race applyed at Idgel Dome (exemple), it was poping the Enter for them, and when entering, they were at the same side, like a group of 12, or sometimes 8 in one instance and 4 on another instance.
The fix ive made, now its validating Arenas and PVP DG, if allow same race, the amount of players need to pop enter.
1 possible problem: After the changes, on Harmony happened to send 1 player of the group (3 members group) to a instance alone, while the rest of the group went to the correct instance against the other group.

Tested on Terath Dredg, Idgel Dome, Kamar, Discipline, Harmony.

SM Dispell:
https://aioncodex.com/5x/skill/3574/?sl=1
This skill, Infernal Pain and Magic Implosion, have a tag (can be dispelled), what it means? if the SM dispell a taget, and the target had only 2 buffs (the magic implosion remove 3 buffs), if the target had only 2 buffs and a debuff like Infernal Pain. The result would be the 2 buffs removed and the Infernal Pain removed. If was 2 buffs and Erosion, the Erosion cannot be removed. and the result would be only 2 buffs of the target removed.

Today its removing even Erosion, so i made a validation to check the STACK/GROUP of only these skills (Infernal Pain, Shackle of Vulnerability and Magic Implosion).

### DIFF
--- a/game-server/src/com/aionemu/gameserver/services/AutoGroupService.java
+++ b/game-server/src/com/aionemu/gameserver/services/AutoGroupService.java
@@ -5,6 +5,7 @@ import java.util.concurrent.ConcurrentHashMap;
 
 import com.aionemu.gameserver.configs.main.AutoGroupConfig;
 import com.aionemu.gameserver.model.ChatType;
+import com.aionemu.gameserver.model.Race;
 import com.aionemu.gameserver.model.autogroup.*;
 import com.aionemu.gameserver.model.gameobjects.player.Player;
 import com.aionemu.gameserver.model.team.alliance.PlayerAllianceService;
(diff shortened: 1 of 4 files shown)


game-server/src/com/aionemu/gameserver/services/AutoGroupService.java
```diff
@@ -53,6 +56,10 @@ public void startLooking(Player player, int maskId, EntryRequestType ert) {
 				PacketSendUtility.broadcastToWorld(
 					new SM_MESSAGE(0, null, player.getRace().getL10n() + " have registered for " + agt.getL10n() + ".", ChatType.BRIGHT_YELLOW_CENTER),
 					p -> p.getRace() != player.getRace() && agt.isInLvlRange(p.getLevel()));
+			}else if( AutoGroupConfig.ANNOUNCE_BATTLEGROUND_REGISTRATIONS && (agt.isHarmonyArena() || agt.isPvPSoloArena()) ){
```
jvbhidalgo, line 59: this can be disregarded, was created to announce applies on Harmony and Discipline.

PR #45, opened 2025-01-29
Fixed quest 1578 not completing correctly

sky1683589933:
Fixed incorrect dialogs.

### DIFF
--- a/game-server/data/handlers/quest/heiron/_1578WhereDoRotronsComeFrom.java
+++ b/game-server/data/handlers/quest/heiron/_1578WhereDoRotronsComeFrom.java
@@ -4,12 +4,10 @@ import static com.aionemu.gameserver.model.DialogAction.*;
 
 import com.aionemu.gameserver.model.gameobjects.Npc;
 import com.aionemu.gameserver.model.gameobjects.player.Player;
-import com.aionemu.gameserver.network.aion.serverpackets.SM_DIALOG_WINDOW;
 import com.aionemu.gameserver.questEngine.handlers.AbstractQuestHandler;
 import com.aionemu.gameserver.questEngine.model.QuestEnv;
 import com.aionemu.gameserver.questEngine.model.QuestState;
 import com.aionemu.gameserver.questEngine.model.QuestStatus;
-import com.aionemu.gameserver.utils.PacketSendUtility;
 
 /**
  * @author MrPoke, Nephis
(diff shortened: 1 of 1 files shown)


game-server/data/handlers/quest/heiron/_1578WhereDoRotronsComeFrom.java
```diff
@@ -58,11 +58,11 @@ else if (env.getDialogActionId() == SETPRO1) {
...
 			if (qs != null && qs.getStatus() == QuestStatus.START && qs.getQuestVarById(0) == 1) {
 				if (env.getDialogActionId() == QUEST_SELECT)
-					return sendQuestDialog(env, 1352);
-				else if (env.getDialogActionId() == SETPRO1) {
+					return sendQuestDialog(env, 1693);
+				else if (env.getDialogActionId() == SETPRO2) {
 					qs.setQuestVarById(0, qs.getQuestVarById(0) + 1);
 					updateQuestStatus(env);
-					PacketSendUtility.sendPacket(player, new SM_DIALOG_WINDOW(env.getVisibleObject().getObjectId(), 1693));
+					PacketSendUtility.sendPacket(player, new SM_DIALOG_WINDOW(env.getVisibleObject().getObjectId(), 10));
```
neon-dev: Quest dialogs should always close after the conversation has finished.
There is a method that advances the quest state and closes the dialog window in a single method. You can replace line 63-66 with:
```java
return defaultCloseDialog(env, 1, 2);
```
1 is the expected quest step (as checked in line 59), 2 is the updated quest step.
Please use this method wherever possible, as it greatly simplifies the handlers.

neon-dev (changes requested):

game-server/data/handlers/quest/heiron/_1578WhereDoRotronsComeFrom.java
sky1683589933: Yes, using  “return defaultCloseDialog (env, 1, 2)”  is more intuitive and concise, but I have found that under some conditions, an empty dialog may appear. Using  “10” to select the next dialogue page will not cause this situation.

neon-dev: The problem you describe could only happen if `defaultCloseDialog()` does not advance the quest step, which is not an expected situation and should be fixed.
Previous dialogs cannot be read again after advancing the quest step and the player should never be able to call SETPRO1, 2, 3, ... twice. Therefore, the problem you have encountered was likely caused by a different problem.

neon-dev: Also, sending dialogPageId 10 is never correct in such cases, as it returns you to the quest selection.
If you want to close the dialog window without advancing the quest step, `return closeDialogWindow(env)` should be used (which sends dialogPageId 0).

sky1683589933: The key is to use  ” updateQuestStatus (env)  “ which can avoid many empy dialog.

neon-dev: `updateQuestStatus(env)` sends the current quest step to the player, but it does not affect the dialog directly.
Only if you combine updating the quest step (setQuestVarById) and sending it (updateQuestStatus), dialogs will be affected, as the handlers will now enter different if..else branches.
Both of these actions also happens inside `defaultCloseDialog()`. You can step into the method in your IDE to see how it works.
Your observations about empty dialogs are based on false assumptions and my statement above still holds true.

Please apply the suggested change here and in all other quests. If you notice problems with empty dialogs, please DM me and we can investigate it together.

sky1683589933: > Also, sending dialogPageId 10 is never correct in such cases, as it returns you to the quest selection. If you want to close the dialog window without advancing the quest step, `return closeDialogWindow(env)` should be used (which sends dialogPageId 0).
Yes, sending "DialogPageId 10" is incorrect no matter what, but it is useful for finding logical errors in the dialogue and determining where the script execution problem lies through client feedback. What would happen if a logical error occurred during script execution using "defaultCloseDialog" directly? There will be no feedback and it will take a lot of time to investigate.

sky1683589933: So I think there is no need to spend time modifying it specifically, because as long as the script logic is written correctly, the result of sending "DialogPageId10" is correct.

neon-dev: `closeDialogWindow` will always close the dialog window. `defaultCloseDialog` will not close the window if the expected quest step is not active. Nothing happening when a dialog option is clicked is still a clear indicator of problems in the quest handler. In my opinion, it's an even clearer feedback to the player that the quest is stuck.
Debugging quest handlers is not difficult once you get the hang of it. Please DM me for debugging tips, I think they will help you a lot.

sky1683589933: Oh, sorry for the writing error! What I want to say is that using 'defaultCloseDialog' when there is an error in the script logic will only get stuck and there will be no feedback from the client (the client will prompt you who to talk to at which step and what to do next).

PR #46, opened 2025-01-30
Fixed quest 1643 cannot be started

sky1683589933:
Fixed incorrect dialogs.


neon-dev (approved):
I have removed the new line that sent the quest dialog page 4 manually, because `sendQuestStartDialog(env, 182201764, 1)` sends the same page on `ASK_QUEST_ACCEPT`.

PR #47, opened 2025-01-30
Fixed quest 2600 can be completed without any items

sky1683589933:
Add complete dialog to resolve dialogue errors, and add conditional judgments to solve the problem of completing the quest without submitting items.


neon-dev (approved):
Very nice work.
I have added another little change to [check the quest item again](https://github.com/beyond-aion/aion-server/pull/47/commits/40c3cce5409c70a74366e7683b973e536127dd00#diff-0a12d6dd5e811e74293865765d66e037d22b94c543a58ecca62abd5bedbea9bdR94) before setting the reward state, in case a dialogue option is skipped (due to an error or exploit).

PR #50, opened 2025-02-04
NPC regular state correction

sky1683589933:
NPC regular state correction


neon-dev (changes requested):
Non-default states are usually only relevant for one spawn position, so it would be better to set them in the spawn templates. Geolus for example, can be set like so in `spawns/Npcs/210030000_Verteron.xml`:
```xml
<!-- Geolus -->
<spawn npc_id="203149" respawn_time="295">
	<spot x="1676.37" y="2270.09" z="166.773" h="35" state="85"/>
</spawn>
```
Please revert all changes in the npc templates and set the states in the individual spawn templates.
Side note: Committing huge files like npc_templates.xml bloats the repository over time, so they should only be changed if necessary.

sky1683589933:
Yeah, submitting a huge template to fix a few irrelevant NPC states is a foolish approach!
Just add it to the spawns, how could I make such a mistake...

PR #52, opened 2025-02-04
Add missing Golden Lepharist Insignia drops in Eltnen

sky1683589933:
Add missing drops for eltnen

### DIFF
--- a/game-server/data/static_data/global_drops/rules/open_worlds/rules_map_eltnen.xml
+++ b/game-server/data/static_data/global_drops/rules/open_worlds/rules_map_eltnen.xml
@@ -1571,6 +1571,7 @@
 			<gd_npc npc_id="211025" /><!-- sorcerer -->
 			<gd_npc npc_id="211029" /><!-- sorcerer -->
 			<gd_npc npc_id="211030" /><!-- lepharist footsoldier -->
+			<gd_npc npc_id="211031" /><!-- lepharist defense commander-->
 		</gd_npcs>
 		<gd_items>
 			<gd_item id="186000046" /><!-- Golden Lepharist Insignia -->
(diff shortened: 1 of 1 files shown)


neon-dev (commented):
I think it would be better to implement the drops similar to [the ones in Heiron](https://github.com/beyond-aion/aion-server/blob/924d53c895ce5a5eae934a21bf254bfd19df8efe/game-server/data/static_data/global_drops/rules/open_worlds/rules_map_heiron.xml#L1648-L1681), if possible.
You can use the `//zone` command to list all zones at your current position.
Please check and let me know what you think.

sky1683589933:
I don't think it's necessary to modify it to group drop. If you want to modify it, you also need to check the accuracy of the group members, because this item is only drop by the NPCs I listed, and there may be other members in this group who do not meet the drop criteria. My idea is that if you want to make modifications, it will take time. Since there are no problems, let's save some time, haha!

game-server/data/static_data/global_drops/rules/open_worlds/rules_map_eltnen.xml
```diff
@@ -1555,4 +1555,25 @@
...
+			<gd_npc npc_id="211021" /><!-- lepharist assaulter -->
+			<gd_npc npc_id="211013" /><!-- lepharist defender -->
+			<gd_npc npc_id="211022" /><!-- lepharist defender -->
+			<gd_npc npc_id="211023" /><!-- assassin -->
+			<gd_npc npc_id="211027" /><!-- assassin -->
+			<gd_npc npc_id="211024" /><!-- lepharist healer -->
+			<gd_npc npc_id="211028" /><!-- lepharist healer -->
+			<gd_npc npc_id="211025" /><!-- sorcerer -->
+			<gd_npc npc_id="211029" /><!-- sorcerer -->
+			<gd_npc npc_id="211030" /><!-- lepharist footsoldier -->
```
neon-dev, line 1573:
```suggestion
			<gd_npc npc_id="211030" /><!-- lepharist footsoldier -->
			<gd_npc npc_id="211031" /><!-- lepharist defense commander-->
```

neon-dev (changes requested):
You are correct that the accuracy must be checked when introducing such changes. I have checked and there are indeed some invalid npcs with `group_drop="LEPHAR"` in the `LEPHARIST_BASTION_210020000` zone. So listing the npc IDs individually is the only correct solution here.

PR #55, opened 2025-02-06
some improvements... optional 😉 

bbortt:
hi there! huge thank you for open sourcing all your code. I've recently installed my own server on a raspberry pi for private usage. it works like a charm 💌 

being a developer myself, I've some remarks on the project. mostly convenience things. I leave it up to you if you merge it all, cherry-pick single commits or don't care at all. this PR just is my way to say "thank you".

- added a Maven wrapper for convenience. not everyone has Maven installed on the local machine (e.g. I don't).
- moved into a monorepo with Maven modules. that makes it possible to conquer all builds simultaneously using `./mvnw package`. if you still want to build single packages, execute `./mvnw package -pl :your-package` (as beforehand). note that this might need adaptions to the `deploy.yml` job, I am not entirely sure.
- updated some dependencies, most notably resolved [`CVE-2024-7254`](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2024-7254) being present in `com.mysql:mysql-connector-j:8.3.0`. I preferrably don't want vulnerabilities in production code.

more to come. if you wish so 😉


commons/pom.xml
```diff
@@ -116,13 +109,13 @@
...
-			<version>8.3.0</version>
+			<version>8.4.0</version>
 		</dependency>
 
 		<dependency>
 			<groupId>org.quartz-scheduler</groupId>
 			<artifactId>quartz</artifactId>
-			<version>2.3.2</version>
+			<version>2.5.0</version>
 			<exclusions>
```
neon-dev: These exclusions are no longer needed with v2.5.0

chat-server/pom.xml
```diff
@@ -2,31 +2,24 @@
...
-				<groupId>org.apache.maven.plugins</groupId>
-				<artifactId>maven-compiler-plugin</artifactId>
-				<version>3.13.0</version>
-				<configuration>
-					<release>21</release>
-					<compilerArgs>
-						<arg>-Xlint:all,-preview,-this-escape</arg>
-					</compilerArgs>
-					<showWarnings>true</showWarnings>
-					<debug>true</debug>
```
neon-dev, line 27: Not sure if removing these parameters in all projects is a good or bad thing. Some of the warnings have been helpful in the past but I can't remember new valid ones in more recent years that I had to fix 🤔

neon-dev (changes requested):
Hi and thank you for your suggestions. At the moment I only have a few notes and one change request, but more may follow:

> not everyone has Maven installed on the local machine

@Estrayl and I haven't decided whether to add the Maven wrapper yet. Personally, I think it's unnecessary, since all the major Java IDEs ship with Maven and I see no reason to support more exotic development environments.

> that makes it possible to conquer all builds simultaneously

Do you mean the servers will build concurrently? Because the parent pom.xml already allows to build everything in one go as it is. Although this is not a feature we use. During development, the IDE takes care of tracking and updating changed classes, and for deployment we do individual builds, as you rarely need to update more than one server.
I'll test and see what practical benefits your changes bring when I find the time, as I haven't really done much with Maven modules yet.

I also have yet to test nightly.yml. Hopefully it has some kind of notification mechanism. If the reports have to be opened manually it is something that can go unnoticed for quite a while. We already have CodeQL, which notifies about vulnerabilities in dependencies, but it doesn't seem to be smart enough to detect them from transitive dependencies.

> more to come. if you wish so 😉

Thanks again. We're always happy when someone decides to contribute instead of keeping improvements and fixes to themselves (which is *very* common in the Aion community 😄).
If you are unsure about future contributions, you can always discuss them beforehand on our Discord.

chat-server/pom.xml
bbortt: funny enough I couldn't build a single module with these parameters. I'll double-check.

bbortt:
thanks for the remarks @neon-dev. I've `cherry-pick`'ed the dependency update commit for now. let me know if you decide to go for the wrapper and/or other changes as well. ~~I personally don't see Maven being shipped with Eclipse Temurin, but I'll gladly be wrong with that one.~~ the advantage of the wrapper is for sure that "works on my machine" does no longer exist - or not because of Maven.

regarding the nightly job; no, there is no notification in place. an alternative for that would be adding an OSSHR upload & report action. those exist. I preferred the manual check up until now.

**edit:** oh, sorry. you spoke about IDE's, not distributions. lol

bbortt:
> Do you mean the servers will build concurrently? Because the parent pom.xml already allows to build everything in one go as it is. Although this is not a feature we use. During development, the IDE takes care of tracking and updating changed classes, and for deployment we do individual builds, as you rarely need to update more than one server.
I'll test and see what practical benefits your changes bring when I find the time, as I haven't really done much with Maven modules yet.

ok, I do see that, "one command to rule them all" is not really an addon. but, using the `parent` for sure has the effect that e.g. the java version must not be declared multiple times. dependencies (-versions) can be managed in one place (e.g. commons module).

individual builds are still possible.

let me know what you think of it, once you tried it. I've now created my ["personal cherries" branch](https://github.com/bbortt/aion-server/tree/feat/personal-cherries), so the PR is more focused on single changes.

.github/workflows/nightly.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Set up JDK 21
+        uses: actions/setup-java@v4
+        with:
+          java-version: '21'
+          distribution: 'temurin'
+          cache: maven
+      - name: Dependency Report
+        run: mvn --no-transfer-progress versions:display-dependency-updates
```
bbortt: removed the Maven wrapper usage, because it is no longer part of this PR.

bbortt:
this should do

commons/pom.xml
```diff
@@ -101,7 +101,7 @@
 		<dependency>
 			<groupId>ch.qos.logback</groupId>
 			<artifactId>logback-classic</artifactId>
-			<version>1.5.6</version>
+			<version>1.5.16</version>
 		</dependency>
 		<dependency>
 			<groupId>com.alibaba.fastjson2</groupId>
```
neon-dev, line 45: This can be updated to 2.0.54 and updating HikariCP to 6.2.1 should be safe too.

.github/workflows/nightly.yml
neon-dev, line 1: I've tried it, and now I remember what I've always disliked about Maven's dependency version analysis: It sometimes recommends release candidates or even alpha versions. It can also select the wrong final version if there are similarly named variants, as in this case with fastjson2: It recommends 2.0.54.android8, when 2.0.54 would be correct. Even after updating to 2.0.54, it still thinks that 2.0.54.android8 is a newer version.
So I think we'll continue to check for new versions the old-fashioned way and you can remove this script.

neon-dev (changes requested):

sky1683589933:
Hi, may I ask if this can solve the issue of inconsistent versions of "slf4j - API" in dependencies?

neon-dev:
> Hi, may I ask if this can solve the issue of inconsistent versions of "slf4j - API" in dependencies?

Well, there is now one less conflict (one instead of two):
grafik

It's not really a problem, the newer version was and still is selected for all the builds. You could define an explicit dependency on it, but with slf4j-api in particular, it's not something to worry about unless there's a vulnerability in the imported version.

sky1683589933:
> It's not really a problem, the newer version was and still is selected for all the builds. You could define an explicit dependency on it, but with slf4j-api in particular, it's not something to worry about unless there's a vulnerability in the imported version.

Yes, I saw the PR regarding this issue and would like to ask by the way.

neon-dev:
> let me know if you decide to go for the wrapper and/or other changes as well.

Hey, sorry for the delay. @Estrayl and I have decided against the Maven wrapper, but the changes to the dependency management and global compiler parameters seem useful.

Just a few last things before I can finally merge your pull request:
1. Please set the properties in the parent pom.xml as follows:
    ```xml
    <properties>
    	<!-- Global -->
    	<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    	<maven.compiler.release>21</maven.compiler.release>
    </properties>
    ```
    The release property is the recommended way to specify the target version since Java 9.
2. Feel free to remove the `showWarnings` and `debug` compiler configs from all pom.xmls, as they are already true by default.
3. We prefer `4.8-SNAPSHOT` over `4.8.0-SNAPSHOT` for this project.
4. I have noticed that `mysql-connector-j` 8.4.0 still depends on the vulnerable protobuf version, but 9.2.0 fixes this problem and looks like a safe upgrade after initial testing and checking the [changelog](https://dev.mysql.com/doc/relnotes/connector-j/en/news-9-x.html).

Thanks again!

.github/workflows/deploy.yml
```diff
@@ -32,4 +32,4 @@ jobs:
       - name: Build with Maven
         env:
           FTP_URL: ${{ secrets.LIVE_SERVER_FTP_URL }}
-        run: mvn -B -Dmaven.source.skip=true -Dmaven.test.skip=${{ github.event.inputs.skip-tests }} -pl :commons,:${{ github.event.inputs.server }} package -Dftp.url=$FTP_URL
\ No newline at end of file
+        run: mvn -B -Dmaven.source.skip=true -Dmaven.test.skip=${{ github.event.inputs.skip-tests }} -pl :${{ github.event.inputs.server }} -am package -Dftp.url=$FTP_URL
```
bbortt, line 35: note this change, removing explicit `:commons` compilation in favour of `-am` (also-make) which builds the parent and commons.

pom.xml
```diff
@@ -13,4 +14,105 @@
...
+			<plugins>
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<artifactId>maven-compiler-plugin</artifactId>
+					<version>3.13.0</version>
+					<configuration>
+						<compilerArgs>
+							<arg>-Xlint:all,-preview,-this-escape</arg>
+						</compilerArgs>
+						<debug>true</debug>
```
bbortt: removed the `showWarnings` ([defaults to `true`](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#showWarnings)), but left the `debug` ([defaults to `false`](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#debug)) arguments.

pom.xml
```diff
@@ -13,4 +14,105 @@
...
+					<version>3.2.5</version>
+				</plugin>
+			</plugins>
+		</pluginManagement>
+
+		<plugins>
+			<plugin>
+				<groupId>io.github.git-commit-id</groupId>
+				<artifactId>git-commit-id-maven-plugin</artifactId>
+				<version>8.0.2</version>
```
bbortt: version `9.0.1` [would be available](https://mvnrepository.com/artifact/io.github.git-commit-id/git-commit-id-maven-plugin), but I haven't checked the changelog. maybe this could be upgraded separately (+1 for small incremental steps from my side 😉).

bbortt (commented):
@neon-dev I've taken your remarks into consideration. please note that I've moved the management of the plugins into the parent `pom.xml`. I hope this is to your liking.

one thing regarding

> 2. Feel free to remove the `showWarngings` and `debug` compiler configs from all `pom.xml`'s, as they are already true by default.

please note my comments.

and regarding

> I have noticed that `mysql-connector-j:8.4.0` still depends on the vulnerable `protobuf `version, but `9.2.0` fixes this problem and looks like a safe upgrade after initial testing and checking the [changelog](https://dev.mysql.com/doc/relnotes/connector-j/en/news-9-x.html).

didn't notice it, and so didn't mvnrepository.com. but nice catch ❤️ I did the upgrade, sounds save based on the changelog.

pom.xml
neon-dev: My IDE tooltip says `debug` defaults to `true`, as well as the documentation here: https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#debug 🤔

bbortt: huh, ok. but then again "set to true to debug" xD that makes no sense. anyway. I'll trust the doc and change it.

neon-dev: Yeah, the description indicates otherwise. Maybe the default value was false in an earlier version 🤷‍♂️

bbortt: did adjust it.

pom.xml
neon-dev: You can include 9.0.1 in this PR, since we're updating pretty much everything at this point.
9.0.1 works for me and writes the version to the manifest as it should. Major breaking change is that [Maven 3.6.3 is now required](https://github.com/git-commit-id/git-commit-id-maven-plugin/releases/tag/v9.0.0), which has been out for over five years now.

bbortt: done.

neon-dev:
During final testing I ran into one problem with this configuration.
I cannot build the servers from their pom.xmls anymore unless I run `mvn install` on the parent pom.xml once. Otherwise I get
```
[ERROR] Failed to read artifact descriptor for com.aionemu:commons:jar:4.8-SNAPSHOT
[ERROR] 	Caused by: The following artifacts could not be resolved: com.aionemu:aion-server:pom:4.8-SNAPSHOT (absent): Could not find artifact com.aionemu:aion-server:pom:4.8-SNAPSHOT
```
Do you know how to solve this without installing the parent pom.xml (+servers) into the local Maven repo?

If there is no other solution, adding this to each server's pom.xmls would at least stop them from being copied into the local Maven repo:
```xml
<properties>
	<maven.install.skip>true</maven.install.skip>
</properties>
```

bbortt:
> I cannot build the servers from their pom.xmls anymore unless I run mvn install on the parent pom.xml once.

can you post the command you're using, please? or do you mean "in IDE" by clicking some "build Maven project" button?

what works is (without previous `mvn install`): `mvn -pl :game-server -am package`. that's [what the deployment job does](https://github.com/beyond-aion/aion-server/pull/55/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3R35), right? nothing is being stored in your local Maven repository.

I could [commit this as a run configuration](https://github.com/bbortt/aion-server/commit/9db08f7ae267e0aa2fd87f8e793643dd21033d0a), if wished?

for any IDE integration to work, you'd have to [enable resolving of workspace artefacts](https://intellij-support.jetbrains.com/hc/en-us/community/posts/5028173549074/comments/5030038280850). I don't not how to do that in other editors, tho. but that requires a previous `mvn package -pl :commons -am` too.

neon-dev:
> can you post the command you're using, please? or do you mean "in IDE" by clicking some "build Maven project" button?

Classic `mvn <phase>` from the server directory, which is the same as double-clicking the phase in IntelliJ's Maven sidebar for example (no run config needed).

> that's [what the deployment job does](https://github.com/beyond-aion/aion-server/pull/55/files#diff-28802fbf11c83a2eee09623fb192785e7ca92a3f40602a517c011b947a1822d3R35), right? nothing is being stored in your local Maven repository.

Yes, because it is run with the parent pom.xml. The module syntax is fine for scripts, but I'd still like to retain the ability to run simple maven phases or goals on the individual module's pom.xmls without too many preconditions.

> I could [commit this as a run configuration](https://github.com/bbortt/aion-server/commit/9db08f7ae267e0aa2fd87f8e793643dd21033d0a), if wished?

Thanks, but I don't think this should be necessary. If this can't be solved inside the pom.xmls in a better way, we should just add the skip property I mentioned earlier to the servers. That way `mvn install` of the parent pom.xml no longer copies the servers to the local Maven repo, even though they would still be compiled and packaged.

bbortt:
> we should just add the skip property I mentioned earlier to the servers

sure thing; I can do that.

neon-dev:
Thanks a lot!

PR #59, opened 2025-02-17
Fixed quest 1393 empty dialogue and stuck issue

sky1683589933:
Added a complete dialog to resolve empty conversations and prevent conversation freezing after interrupting, and added a complete area exploration event that is closer to reality.


neon-dev:
I removed custom handling for `SELECT1_1_1` because the default handling responds with the same dialogue already.

PR #60, opened 2025-02-17
Fixed Quest 4210 conversation error

sky1683589933:
Modify dialogue conditions based on NPC type.

### DIFF
--- a/game-server/data/handlers/quest/beluslan/_4200ASuspiciousCall.java
+++ b/game-server/data/handlers/quest/beluslan/_4200ASuspiciousCall.java
@@ -74,18 +74,16 @@ public class _4200ASuspiciousCall extends AbstractQuestHandler {
 			}
 			return false;
 		} else if (qs.getStatus() == QuestStatus.START) {
-			if (targetId == 204839)// Uikinerk
-			{
+			if (targetId == 204839 && var == 0) { // Uikinerk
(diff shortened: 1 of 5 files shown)


neon-dev:
Based on the dialogue I've found that the npc was the wrong one and pushed a fix for all affected quests. Please test if everything works as expected.

sky1683589933:
> Based on the dialogue I've found that the npc was the wrong one and pushed a fix for all affected quests. Please test if everything works as expected.

Okay, we'll test it after we get home tonight
It can be confirmed that once you have made the modifications, there will be no issues. I just want to see if any quests associated with the NPC changes will be affected.

sky1683589933:
> Based on the dialogue I've found that the npc was the wrong one and pushed a fix for all affected quests. Please test if everything works as expected.

Yeah, using NPC (798332) in quest 4210 is incorrect.
I didn't carefully examine the details of the dialog, and the online database is also incorrect: https://aioncodex.com/4x/quest/4210/?sl=1 .

sky1683589933:
I checked and found that there were no quest affected by the deletion of NPC (798332).
The "Haorunerk" I saw through the viewing dialog should be a corpse, and the living "Haorunerk" in quest 4200 is just a shadow of the appearance of the timeline ahead of time.
Fixed very well!

PR #61, opened 2025-02-17
Fixed quest 1636

sky1683589933:
Fixed the use of zones for items in quest 1636 and added a spawn world boss(Based on retail).


neon-dev:
I have changed some things that were not retail-like:
- fixed that the statue could not be used (and removed the zone check, as the zone is smaller than the statue's interaction range)
- the statue is now temporarily deleted after use
- the flute can no longer be used directly (it has no item action and the quest description tells you to use the statue first)
- added missing drops for Akairun of Medeus

sky1683589933:
> I have changed some things that were not retail-like:
> 
> * fixed that the statue could not be used (and removed the zone check, as the zone is smaller than the statue's interaction range)
> * the statue is now temporarily deleted after use
> * the flute can no longer be used directly (it has no item action and the quest description tells you to use the statue first)
> * added missing drops for Akairun of Medeus

I think removing the restriction on the use zone of items is possible, but it should be kept that only clicking on the statue while holding the flute (182201785) can complete the quest.

sky1683589933:
When I was testing, I had to jump onto the base of the statue to meet the usage conditions, but there were some issues with the zone.

neon-dev:
> it should be kept that only clicking on the statue while holding the flute (182201785) can complete the quest.

I think the statue should only check for the flute item if the quest can automatically reset the quest when discarding the item, so the player is not stuck in an unrecoverable quest state.
This could be implemented more easily once there is something like a `onWorkItemDiscarded` method.
I am also not really sure if every quest is supposed to reset the step if a work item is deleted. Maybe we can find some info from official servers soon.

> When I was testing, I had to jump onto the base of the statue to meet the usage conditions, but there were some issues with the zone.

This is the same problem I have experienced, which is why I have removed the zone check. The zone doesn't extend past the base of the statue, but the quest must be able to progress without jumping on the base.

sky1683589933:
Let me test it first, haha!

sky1683589933:
> This is the same problem I have experienced, which is why I have removed the zone check. The zone doesn't extend past the base of the statue, but the quest must be able to progress without jumping on the base.

The "LF3_LIMITAREA_F4_210040000" zone covers the entire upper disc, and I just used it for testing and the results were very good.

neon-dev:
> The "LF3_LIMITAREA_F4_210040000" zone covers the entire upper disc, and I just used it for testing and the results were very good.

This zone is not related to the quest and is not required to check, because the statue exists only in one place and has a fixed interaction range.

game-server/data/static_data/global_drops/rules/open_worlds/rules_map_heiron.xml
```diff
@@ -393,6 +409,21 @@
 			<gd_item id="100500445"/><!-- Medeus' Heart -->
 		</gd_items>
 	</gd_rule>
+	<gd_rule rule_name="Named Weapon (Unique)" chance="100">
+		<gd_maps>
+			<gd_map map_id="210040000"/><!-- Heiron -->
+		</gd_maps>
+		<gd_npcs>
+			<gd_npc npc_id="212008"/><!-- Akairun of Medeus -->
```
sky1683589933, line 417:
```suggestion
			<gd_npc npc_id="290106"/><!-- Akairun of Medeus -->
```

game-server/data/static_data/global_drops/rules/open_worlds/rules_map_heiron.xml
```diff
@@ -365,6 +365,22 @@
 			<gd_item id="100100524"/><!-- Ursanafi's Warhammer -->
 		</gd_items>
 	</gd_rule>
+	<gd_rule rule_name="Named Weapon (Legend)" chance="100">
+		<gd_maps>
+			<gd_map map_id="210040000"/><!-- Heiron -->
+		</gd_maps>
+		<gd_npcs>
+			<gd_npc npc_id="212008"/><!-- Akairun of Medeus -->
```
sky1683589933, line 373:
```suggestion
			<gd_npc npc_id="290106"/><!-- Akairun of Medeus -->
```

game-server/data/static_data/global_drops/rules/open_worlds/rules_map_heiron.xml
```diff
@@ -1162,6 +1193,18 @@
 			<gd_item id="122000795"/><!-- Medeus' Ring -->
 		</gd_items>
 	</gd_rule>
+	<gd_rule rule_name="Named Accessory (Legend)" chance="100">
+		<gd_maps>
+			<gd_map map_id="210040000"/><!-- Heiron -->
+		</gd_maps>
+		<gd_npcs>
+			<gd_npc npc_id="212008"/><!-- Akairun of Medeus -->
```
sky1683589933, line 1201:
```suggestion
			<gd_npc npc_id="290106"/><!-- Akairun of Medeus -->
```

game-server/data/static_data/global_drops/rules/open_worlds/rules_map_heiron.xml
```diff
@@ -1364,6 +1407,20 @@
 			<gd_item id="125001565"/><!-- Viper's Mask -->
 		</gd_items>
 	</gd_rule>
+	<gd_rule rule_name="Named Headband (Legend)" chance="100">
+		<gd_maps>
+			<gd_map map_id="210040000"/><!-- Heiron -->
+		</gd_maps>
+		<gd_npcs>
+			<gd_npc npc_id="212008"/><!-- Akairun of Medeus -->
```
sky1683589933, line 1415:
```suggestion
			<gd_npc npc_id="290106"/><!-- Akairun of Medeus -->
```

game-server/data/static_data/global_drops/rules/open_worlds/rules_map_heiron.xml
```diff
@@ -1526,6 +1583,29 @@
 			<gd_item id="152012031"/><!-- Sentinel's Branch -->
 		</gd_items>
 	</gd_rule>
+	<gd_rule rule_name="Named Material" chance="100">
+		<gd_maps>
+			<gd_map map_id="210040000"/><!-- Heiron -->
+		</gd_maps>
+		<gd_npcs>
+			<gd_npc npc_id="212008"/><!-- Akairun of Medeus -->
```
sky1683589933, line 1591:
```suggestion
			<gd_npc npc_id="290106"/><!-- Akairun of Medeus -->
```

game-server/data/static_data/global_drops/rules/open_worlds/rules_map_heiron.xml
```diff
@@ -1526,6 +1583,29 @@
...
+		<gd_items>
+			<gd_item id="152012033"/><!-- Akairun's Horn -->
+		</gd_items>
+	</gd_rule>
+	<gd_rule rule_name="Named Design" chance="100">
+		<gd_maps>
+			<gd_map map_id="210040000"/><!-- Heiron -->
+		</gd_maps>
+		<gd_npcs>
+			<gd_npc npc_id="212008"/><!-- Akairun of Medeus -->
```
sky1683589933:
```suggestion
			<gd_npc npc_id="290106"/><!-- Akairun of Medeus -->
```

game-server/data/handlers/quest/heiron/_1636AFluteForTheFixing.java
```diff
@@ -66,6 +65,12 @@ public boolean onDialogEvent(QuestEnv env) {
 						case FINISH_DIALOG:
 							return sendQuestSelectionDialog(env);
 					}
+					break;
+				case 700239: // Drake Stone Statue
+					playQuestMovie(env, 210);
+					VisibleObject drakeStoneStatue = env.getVisibleObject();
+					spawnTemporarily(212008, player.getWorldMapInstance(), drakeStoneStatue.getX(), drakeStoneStatue.getY(), drakeStoneStatue.getZ(), (byte) 30, 60);
```
sky1683589933:
```suggestion
					spawnTemporarily(290106, player.getWorldMapInstance(), drakeStoneStatue.getX(), drakeStoneStatue.getY(), drakeStoneStatue.getZ(), (byte) 30, 60);
```

sky1683589933 (commented):
The final weakened version after changing the boss to 4.0.

game-server/data/static_data/global_drops/rules/open_worlds/rules_map_heiron.xml
```diff
@@ -1594,6 +1594,18 @@
 			<gd_item id="152012033"/><!-- Akairun's Horn -->
 		</gd_items>
 	</gd_rule>
+	<gd_rule rule_name="Named Design" chance="100">
```
neon-dev: These designs are not supposed to drop. They can be bought from npcs such as https://aioncodex.com/5x/npc/802413

neon-dev:
I am not sure about the npc ID you changed.
Data from the 4.6 PTS templates:
212008 has `<__spawn_zonename__>LF3</__spawn_zonename__>` (Heiron)
290017 has `<__spawn_zonename__>ab1_1211</__spawn_zonename__>` (Reshanta - Roah Fortress)
290106 has `<__spawn_zonename__>ab1_1011</__spawn_zonename__>` (Reshanta - Divine Fortress)

I can also find no drops for 290017 and 290106 in the 4.6 PTS files.

sky1683589933:
> I am not sure about the npc ID you changed. Data from the 4.6 PTS templates: 212008 has `<__spawn_zonename__>LF3</__spawn_zonename__>` (Heiron) 290017 has `<__spawn_zonename__>ab1_1211</__spawn_zonename__>` (Reshanta - Roah Fortress) 290106 has `<__spawn_zonename__>ab1_1011</__spawn_zonename__>` (Reshanta - Divine Fortress)
> 
> I can also find no drops for 290017 and 290106 in the 4.6 PTS files.

Okay, I was too subjective!
It can be confirmed that these two IDs (290017, 290106) were not enabled before 6.0. I would like to say that sometimes subjectivity may be better than the official setting, because NC setting of bosses in this world is extremely unfair. In the official server, this boss is difficult to fight, but the properties of dropped items are not good. If according to the official setting, no one will fight this boss, losing its purpose of existence.

game-server/data/static_data/global_drops/rules/open_worlds/rules_map_heiron.xml
sky1683589933: > These designs are not supposed to drop. They can be bought from npcs such as https://aioncodex.com/5x/npc/802413

I think it is still necessary to add the production drawings. In my opinion, the official setting is meaningless as it requires purchasing from the Koldor map. If you reach level 65, who would care about creating a level 45 item. What's the point at that time!

neon-dev: This was just an example. You don't have to be level 65 to buy the designs, which would be pointless indeed.
There are more npcs selling it, even one in Sanctum, according to the "Sold by NPC" tab here: https://aioncodex.com/5x/item/152200508/?sl=1
I also found one in Divine Fortress that is not listed on Aioncodex (don't remember the ID right now).

sky1683589933: I couldn't find any NPCs selling this production drawing in the 4.8 game. What I want to say is that there are some methods that exist in 5. x but not in 4.8. Although NPCs have been made, they haven't been used yet. 4.8 is a major transformation version, and using a single query for 4. x is somewhat inappropriate, and using a single query for 5. x is also somewhat inappropriate.

neon-dev: Again, this was just an example. I originally searched the 4.8 client data for the merchants and found multiple per design. I did not fully validate which ones should be spawned, as it was just a very quick initial search. So I just mentioned the first one I could find an existing spawn position in the emulator for.
In my opinion, if there are merchants that can be accessed at level <=50, custom drops should not be added.
I can do a complete search in the client files and PTS files later.

neon-dev:
> I would like to say that sometimes subjectivity may be better than the official setting

I totally agree.

> In the official server, this boss is difficult to fight, but the properties of dropped items are not good. If according to the official setting, no one will fight this boss, losing its purpose of existence.

Most world bosses have become irrelevant beyond version 3.x, which is sad. Modifying one secret boss that 99.x% of players will never meet will not solve this. Server owners can choose if and how they want to modify them on a global level to make them more attractive.
Additionally, 290106 has incorrect stats in the emulator. But even after fixing, it would be still too weak and low level (40) to drop unique level 46 weapons, especially at a 100% drop rate. It is not a good substitute. The way NCsoft has nerfed the game with later versions was often too extreme and is not something I always agree with.
For these reasons I find it difficult to justify this custom boss ID here. Since it can only be spawned once, I prefer the official hard boss with (custom) 100% drop chances over a weak and meaningless one who might not drop unique weapons at all.

Like I said, server owners can always change the game to make it more to their liking. This is a highly subjective matter, so there are no right or wrong answers.

sky1683589933:
It has been confirmed that the summoning boss (ID: 212008) in quest 1636 clicked on the statue instead of the flute, and the repaired flute (182201785) was a condition for advancing the quest and summoning the boss.
The two design patterns also have drops.

neon-dev:
Thanks for your tests and further corrections.

I have checked with PTS drops and found that both designs are world drops in Heiron (and Theobomos). They are part of several large design drop groups. I only implemented drops for these two designs in Heiron to keep it simple. It would be too much work to correctly implement all design drops from the PTS drop groups on this map.

I also removed the item check you have added when interacting with the statue, because the quest can not recover from a state where the player discards the flute. The `onCanAct()` method checks that the statue can only be used in the correct quest step, so everything is working without problems now.
If you want to check for the flute, I think throwing it away should not only only stop the quest from progressing, but abandon the quest completely. This would be a logical consequence for this quest, as the flute is an essential item that can not be recovered from an earlier quest step.

sky1683589933:
> If you want to check for the flute, I think throwing it away should not only only stop the quest from progressing, but abandon the quest completely. This would be a logical consequence for this quest, as the flute is an essential item that can not be recovered from an earlier quest step.

The added inventory detection is only to make up for the insufficient functionality of the "quest_use_item" AI in all emulators. It is this AI that really needs to be fixed, so deleting inventory detection is valid.

This way, there are no logical issues with the execution script, that's great!

neon-dev:
The item can be checked in `onCanAct()` to skip the usage animation on the statue, if that is the insufficient functionality you were thinking about.
But adding this item check in `onCanAct()` would introduce the same problem of the quest getting stuck if the player threw away the flute.

sky1683589933:
What I mean by insufficient AI functionality is that it cannot detect whether there are registered work items in the PC backpack, for example, if you click on a locked door (NPC), it needs to detect whether there is a corresponding key in your backpack.

neon-dev:
The presence of work items can be checked in `onCanAct()`, like I just mentioned. If the method returns false, the player cannot interact with the npc.

sky1683589933:
> The presence of work items can be checked in `onCanAct()`, like I just mentioned. If the method returns false, the player cannot interact with the npc.

What I mean is that if the player loses the 'repaired flute', nothing will continue, the boss cannot summon and the quest cannot continue, they can only give up and start over.
Simply put, there is no repaired flute in the backpack "case 700239 return false".

sky1683589933:
I think this is also the reason why the author who wrote this execution script used 'useItemEvent' to bypass the problem with AI itself, but this has resulted in many logic that does not match reality.

sky1683589933:
I think there is no problem with the current fix, because the use of the statue is part of the Quest steps and can only be used in specific step states without checking whether the "repaired flute" is held, so there will be no logical loopholes. The detection work item I proposed is only based on the inference of dialogue logic and is not necessary，There is indeed no response to lost work items in Quest.

neon-dev:
I agree. We could make another improvement and spawn the boss after the movie has finished, so it doesn't attack you during the cutscene.
What do you think?

sky1683589933:
> I agree. We could make another improvement and spawn the boss after the movie has finished, so it doesn't attack you during the cutscene. What do you think?

I can be certain that in the official settings, if players focus on watching the animation and wait for it to end, they may be instantly killed by the boss, haha!
But I don't think there's anything wrong with the changes you mentioned!

neon-dev:
I've revised quest movie events so it's easier to work with them (see 94beb4a90c002571ecf0851a19dff09a94a24ab1). This commit was merged just before f727aae, so `env.getVisibleObject()` can be used in `onMovieEndEvent()`.
Finally merged this pull request, as it's now working well enough. Thank you for all your hard work and feedback!

PR #80, opened 2025-08-09
Fix: Corrected AI behavior for the "Eternal Bastion entrance" portal to make it functional.

jeanmarcos-dev:
The portal was previously using an incorrect AI (aggressive) which prevented interaction. Updated it to use the proper portal_dialog AI, allowing players to use it as intended to access the Steel Wall Bastion corridor.

image


neon-dev:
Thanks!
I found and fixed more broken portals with this PR, to avoid multiple commits with this huge template file (keeps the repo smaller).

PR #81, opened 2025-08-15
Allow opening Private Store with active Power Shards

jeanmarcos-dev:
This change adds a new ACTIVE_WITH_POWERSHARD state to the CreatureState enum, combining the active state with the use of Power Shards.
In PrivateStoreService, the canOpenPrivateStore validation is updated to allow players to open their private store when in either ACTIVE or ACTIVE_WITH_POWERSHARD state.

Main Changes:

Added ACTIVE_WITH_POWERSHARD state in CreatureState.java.

Updated canOpenPrivateStore condition to also accept ACTIVE_WITH_POWERSHARD.

Motivation:
Previously, players were unable to open the private store if they had Power Shards active. This change removes that unnecessary restriction while keeping compatibility with the standard active state.


neon-dev:
Thanks for the suggested changes. However, I have identified some more problematic cases and found a cleaner solution that resolves all of them.

PR #82, opened 2025-08-15
Refactor NPC dialog handling and quest event processing

jeanmarcos-dev:
# PR Title
Refactor NPC dialog handling and fix "load fail!" error on certain NPCs

## Description
This update simplifies and streamlines NPC dialog interaction logic, removes unnecessary calls to `QuestEngine.onDialog`, and consolidates duplicated quest handling code.  
Additionally, it fixes the issue where certain NPCs (e.g., *Ancient Crown Administration Officer*, *Ancient Goblet Administration Officer*) displayed a `load fail!` message when interacted with.

## Main Changes
- **`NpcController`**:  
  - Removed redundant `QuestEngine.onDialog` call when out of range.  
  - Simplified position check with an early return if the player is not in talk range.  
  - Retained AI dialog handling and ensured consistent call to `DialogService.onDialogSelect`.

- **`AbstractQuestHandler`**:  
  - Replaced `sendDialogPacket` with a direct `PacketSendUtility.sendPacket` call to `SM_DIALOG_WINDOW`.

- **Dialog handling refactor**:  
  - Introduced `handleQuestDialog` helper method to centralize quest environment creation and processing.  
  - Updated multiple dialog interaction flows to use the new helper method, removing duplicated `QuestEnv` creation.  
  - Ensured consistent handling of `extendedRewardIndex` in all dialog paths.

## Motivation

image

Previously, some NPCs triggered a `load fail!` error due to redundant or incorrect dialog event handling.  
This refactor removes unnecessary calls, ensures consistent quest dialog processing, and fixes the dialog display for affected NPCs, allowing their HTML pages to load properly without breaking interactions.

### DIFF
--- a/game-server/src/com/aionemu/gameserver/controllers/NpcController.java
+++ b/game-server/src/com/aionemu/gameserver/controllers/NpcController.java
@@ -263,14 +263,10 @@ public class NpcController extends CreatureController<Npc> {
 
 	@Override
 	public void onDialogSelect(int dialogActionId, int prevDialogId, Player player, int questId, int extendedRewardIndex) {
-		if (!PositionUtil.isInTalkRange(player, getOwner())) {
-			// TODO remove this call and return unconditionally once logging has clarified wtf it is trying to solve 
-			if (QuestEngine.getInstance().onDialog(new QuestEnv(getOwner(), player, questId, dialogActionId)))
-				log.warn(player + " talked to " + getOwner() + " out of range, but QuestEngine.onDialog returned true");
-			else
-				return;
-		}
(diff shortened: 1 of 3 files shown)


neon-dev (approved):
I was worried that your changes in `AbstractQuestHandler#onDialogEvent()` might cause issues with certain quests, but during testing I found that the fallback handling in the DialogService catches them all. I could imagine theoretical cases where this change could break a quest, but I'm not aware of quests that meet the specific criteria for this to happen.
Therefore, I don't think merging this pull request should cause any major issues. Finally, I cleaned up your commit to remove some unnecessary code.

PR #85, opened 2025-11-28
Fix //set gp admin command

sined-dev:
Correction of the admin command //set gp in cases where the specified value is below zero.


neon-dev:
Contrary to the chat command's name, it never set AP or GP to the specified amount. f9e3764 fixes this and a few other issues.

PR #90, opened 2025-12-13
Fixed pull mechanics as on retail

crlfrancis:
Pulls should work through damage reduction shields which still does damage


neon-dev:
This was changed in d75b4d4. @Estrayl will have to check.

Estrayl (approved):
Verified on 4.6 PTS:
If the shield is broken, the pull goes through.
If the shield is not broken, the pull is also resisted.

neon-dev:
This fixes shields like Iron Skin, which reduce the damage by 50%. Shields like Stone Skin, which reduce damage by 100% were not affected.

PR #97, opened 2025-12-16
Fix movement modifiers applied to unsupported damage effects

Birthright50:
### Summary
Fixes incorrect application of movement-based damage modifiers.

### Details
Movement modifiers are now applied only to damage effects that explicitly
support movement-based scaling. Effects such as `DispelBuffCounterATK`, `DelayedSpellATK_Instant` , `ProcATK_Instant`  `NoReduceSpellATK_Instant`  are correctly excluded. (There may actually be more effects.) Tested on 4.6 PTS.

This change addresses a limitation mentioned in a previous pull request:
[Fix elemental defend while moving: constant value instead percent](https://github.com/beyond-aion/aion-server/pull/93)

### DIFF
--- a/game-server/src/com/aionemu/gameserver/controllers/attack/AttackUtil.java
+++ b/game-server/src/com/aionemu/gameserver/controllers/attack/AttackUtil.java
@@ -314,7 +314,7 @@ public class AttackUtil {
 				damageMultiplier = shouldIncreaseByOneTimeBoost ? effector.getObserveController().getBaseMagicalDamageMultiplier() : 1f;
 				damage = StatFunctions.calculateMagicalSkillDamage(effector, effected, damage, (int) bonus, element, true, true);
 			}
-			if (template.shouldApplyAttackerMovementModifier()){
+			if (template.shouldApplyAttackerMovementModifier()) {
 				damage = StatFunctions.adjustStatByMovementModifier(effector, isPhysical ? StatEnum.PHYSICAL_ATTACK : StatEnum.MAGICAL_ATTACK, damage);
 			}
 			damage *= damageMultiplier;
--- a/game-server/src/com/aionemu/gameserver/controllers/attack/AttackUtil.java
+++ b/game-server/src/com/aionemu/gameserver/controllers/attack/AttackUtil.java
@@ -230,7 +230,6 @@ public class AttackUtil {
 		// define values
 		ActionModifier modifier = template.getActionModifiers(effect);
 		SkillElement element = template.getElement();
-		Func func = template.getMode();
 		int randomDamageType = template instanceof SkillAttackInstantEffect skillAttackInstantEffect ? skillAttackInstantEffect.getRnddmg() : 0;
 		int critAddDmg = template.getCritAddDmg2() + template.getCritAddDmg1() * effect.getSkillLevel();
 		boolean useTemplateDmg = isUseTemplateDmg(effect, template);
@@ -289,11 +288,9 @@ public class AttackUtil {
 			damage += res.getExactDamage();
 		}
 		// add skill damage
-		if (func != null) {
-			switch (func) {
-				case ADD -> damage += skillDamage;
-				case PERCENT -> damage += baseAttack * skillDamage / 100f;
-			}
+		switch (template.getMode()) {
+			case ADD -> damage += skillDamage;
+			case PERCENT -> damage += baseAttack * skillDamage / 100f;
 		}
 
 		// add bonus damage
--- a/game-server/src/com/aionemu/gameserver/controllers/attack/AttackUtil.java
+++ b/game-server/src/com/aionemu/gameserver/controllers/attack/AttackUtil.java
@@ -224,13 +224,13 @@ public class AttackUtil {
 		};
 	}
 
-	public static void calculateSkillResult(Effect effect, int skillDamage, EffectTemplate template, boolean ignoreShield) {
+	public static void calculateSkillResult(Effect effect, int skillDamage, DamageEffect template, boolean ignoreShield) {
 		Creature effector = effect.getEffector();
 		Creature effected = effect.getEffected();
 		// define values
 		ActionModifier modifier = template.getActionModifiers(effect);
 		SkillElement element = template.getElement();
-		Func func = template instanceof DamageEffect damageEffect ? damageEffect.getMode() : Func.ADD;
+		Func func = template.getMode();
 		int randomDamageType = template instanceof SkillAttackInstantEffect skillAttackInstantEffect ? skillAttackInstantEffect.getRnddmg() : 0;
 		int critAddDmg = template.getCritAddDmg2() + template.getCritAddDmg1() * effect.getSkillLevel();
 		boolean useTemplateDmg = isUseTemplateDmg(effect, template);
@@ -318,7 +318,9 @@ public class AttackUtil {
 				damageMultiplier = shouldIncreaseByOneTimeBoost ? effector.getObserveController().getBaseMagicalDamageMultiplier() : 1f;
 				damage = StatFunctions.calculateMagicalSkillDamage(effector, effected, damage, (int) bonus, element, true, true);
 			}
-			damage = StatFunctions.adjustStatByMovementModifier(effector, isPhysical ? StatEnum.PHYSICAL_ATTACK : StatEnum.MAGICAL_ATTACK, damage);
+			if (template.shouldApplyMovementModifier()){
+				damage = StatFunctions.adjustStatByMovementModifier(effector, isPhysical ? StatEnum.PHYSICAL_ATTACK : StatEnum.MAGICAL_ATTACK, damage);
+			}
 			damage *= damageMultiplier;
 		}
 
--- a/game-server/src/com/aionemu/gameserver/controllers/attack/AttackUtil.java
+++ b/game-server/src/com/aionemu/gameserver/controllers/attack/AttackUtil.java
@@ -19,7 +19,6 @@ import com.aionemu.gameserver.model.templates.item.ItemAttackType;
 import com.aionemu.gameserver.model.templates.item.enums.ItemGroup;
 import com.aionemu.gameserver.model.templates.item.enums.ItemSubType;
 import com.aionemu.gameserver.model.templates.npc.NpcTemplate;
-import com.aionemu.gameserver.skillengine.change.Func;
 import com.aionemu.gameserver.skillengine.effect.*;
 import com.aionemu.gameserver.skillengine.effect.modifier.ActionModifier;
 import com.aionemu.gameserver.skillengine.model.Effect;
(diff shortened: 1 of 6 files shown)


game-server/src/com/aionemu/gameserver/controllers/attack/AttackUtil.java
```diff
@@ -318,7 +318,9 @@ public static void calculateSkillResult(Effect effect, int skillDamage, EffectTe
 				damageMultiplier = shouldIncreaseByOneTimeBoost ? effector.getObserveController().getBaseMagicalDamageMultiplier() : 1f;
 				damage = StatFunctions.calculateMagicalSkillDamage(effector, effected, damage, (int) bonus, element, true, true);
 			}
-			damage = StatFunctions.adjustStatByMovementModifier(effector, isPhysical ? StatEnum.PHYSICAL_ATTACK : StatEnum.MAGICAL_ATTACK, damage);
+			if (template.shouldApplyMovementModifier()){
```
neon-dev:
```suggestion
			if (template.shouldApplyMovementModifier()) {
```

game-server/src/com/aionemu/gameserver/controllers/attack/AttackUtil.java
```diff
@@ -224,13 +224,13 @@ private static float getWeaponMultiplier(ItemGroup group) {
...
 
-	public static void calculateSkillResult(Effect effect, int skillDamage, EffectTemplate template, boolean ignoreShield) {
+	public static void calculateSkillResult(Effect effect, int skillDamage, DamageEffect template, boolean ignoreShield) {
 		Creature effector = effect.getEffector();
 		Creature effected = effect.getEffected();
 		// define values
 		ActionModifier modifier = template.getActionModifiers(effect);
 		SkillElement element = template.getElement();
-		Func func = template instanceof DamageEffect damageEffect ? damageEffect.getMode() : Func.ADD;
+		Func func = template.getMode();
```
neon-dev: I'd remove this here and the null check on line 292. Can be used in the switch statement on line 293 without a null check now.

game-server/src/com/aionemu/gameserver/skillengine/effect/DispelBuffCounterAtkEffect.java
```diff
@@ -46,6 +46,11 @@ public void calculateDamage(Effect effect) {
 		AttackUtil.calculateSkillResult(effect, valueWithDelta, this, false);
 	}
 
+	@Override
+	public boolean shouldApplyMovementModifier() {
+		return false;
+	}
+
```
neon-dev, line 53: In my 4.6 PTS tests, the movement modifier always applied (tested with Ignite Aether I).

game-server/src/com/aionemu/gameserver/skillengine/effect/DelayedSpellAttackInstantEffect.java
```diff
@@ -40,4 +40,9 @@ public void applyEffect(Effect effect) {
 	@Override
 	public void calculateDamage(Effect effect) {
 	}
+
+	@Override
+	public boolean shouldApplyMovementModifier() {
+		return false;
+	}
```
neon-dev: In my 4.6 PTS tests, the movement modifier applied if the target was moving on cast end (tested with Delayed Blast I), but didn't apply if the target only started moving after the skill had already been cast.

game-server/src/com/aionemu/gameserver/skillengine/effect/ProcAtkInstantEffect.java
```diff
@@ -21,6 +21,11 @@ public void applyEffect(Effect effect) {
 		effect.getEffected().getController().onAttack(effect, TYPE.DAMAGE, damage, true, LOG.PROCATKINSTANT, hopType);
 	}
 
+	@Override
+	public boolean shouldApplyMovementModifier() {
+		return false;
+	}
+
```
neon-dev, line 28: In my 4.6 PTS tests, the movement modifier always applied (tested with Godstone: Zikel's Outrage).

neon-dev (changes requested):

game-server/src/com/aionemu/gameserver/skillengine/effect/DispelBuffCounterAtkEffect.java
Birthright50: https://www.youtube.com/watch?v=GnujfoC0Wdk 4.6
https://www.youtube.com/watch?v=mfnrWWZB_wo classic RuOff

game-server/src/com/aionemu/gameserver/skillengine/effect/ProcAtkInstantEffect.java
Birthright50: https://www.youtube.com/watch?v=-g9vI4Q2Xt8 ProcATK_Instant Godstone 10% 376

game-server/src/com/aionemu/gameserver/skillengine/effect/DelayedSpellAttackInstantEffect.java
Birthright50: Target or attacker? 
https://www.youtube.com/watch?v=5qxJk12XJJI 
Yes, you’re right — this is my mistake. Movement bonuses do apply to
`DelayedSpellATK_Instant` effects.

However, there is an important nuance worth mentioning.

On retail servers, movement-based bonuses are applied with a delay:
they do not activate immediately when movement starts, and they also
do not disappear instantly when movement stops. There is a short grace
period in both cases.

On Java-based servers, this behavior is not implemented correctly.
Movement bonuses are applied immediately as soon as the attacker starts
moving and are removed instantly when movement stops. Because of this,
movement bonuses are effectively unreliable or even impossible to apply
correctly to cast-based skills on Java servers.

Given this, I have a question regarding the intended logic here:
should `applyEffect` (and damage calculation for delayed spells) use
the attacker’s movement state at the moment the cast finishes,
or the movement state at the moment the delayed damage is actually applied?

Clarifying this would help determine the correct behavior for
`DelayedSpellATK_Instant`.

neon-dev: I have missed to test the attackers modifiers. So far, I have only tested the (defensive) modifiers of a a moving target with a stationary attacker.
So all my previous comments were just related to one part of the retail behavior. It looks like the attacker's offensive modifiers are handled differently than the target's defensive modifiers, which needs more testing.

> On Java-based servers, this behavior is not implemented correctly.

Some of the problems have been fixed in 19f9904.

> they do not activate immediately

I believe this is the only remaining problem regarding activation of movement modifiers.

> Given this, I have a question regarding the intended logic here:
should applyEffect (and damage calculation for delayed spells) use
the attacker’s movement state at the moment the cast finishes,
or the movement state at the moment the delayed damage is actually applied?

Since I have not yet tested the attacker's modifiers, there is no clear answer to this question. The target's modifiers seem to be apply right when the cast finishes, but we should also identify the exact delay for modifier activation and deactivation to make sure we know when the modifiers are actually active.

Birthright50: > I have missed to test the attackers modifiers. So far, I have only tested the (defensive) modifiers of a a moving target with a stationary attacker.

I think we are actually looking at two different
movement-related mechanics here, which might be causing the confusion.

However, the changes in this PR affect **attacker-side movement modifiers only**.

More specifically:
- This PR does **not** change defensive modifiers of the target.
- It only affects whether **offensive movement bonuses of the attacker**
  are applicable to certain damage effect types.

For reference, on retail:
- When a target moves forward, elemental defense is reduced by a flat value
  (e.g. -50), but this reduction **does not apply to DoT effects**
  (`SpellATK` interval-based damage). Its OK now.
- This behavior is separate from attacker movement bonuses.



> Since I have not yet tested the attacker's modifiers, there is no clear answer to this question. The target's modifiers seem to be apply right when the cast finishes, but we should also identify the exact delay for modifier activation and deactivation to make sure we know when the modifiers are actually active.

One additional clarification based on the attached videos:

On retail, movement-based modifiers for delayed skills are evaluated
at the moment the cast finishes, not at the moment the delayed damage
is actually applied.

The same timing applies to defender-side modifiers:
- If the target is moving forward when the cast completes, elemental
  defense reduction applies even if the target stops moving afterward.
- If the cast completes while the target is stationary, elemental
  defense reduction does not apply, even if the target starts moving
  forward after that.

This indicates that movement state is effectively snapshotted
at cast completion.

> I believe this is the only remaining problem regarding activation of movement modifiers.

Regarding the remaining issue with movement modifiers activating and deactivating
with a delay on retail: at the moment it is unclear how this delay can be
reliably derived.

On the client side, movement bonuses are indicated visually (arrows for forward,
side, or backward movement), but without access to client source code it is
difficult to determine the exact timing logic behind these state changes.
Estimating the delay purely from video footage is also imprecise.

A similar problem exists with weapon swapping: on Java-based servers weapon swaps
are effectively instant, while on retail the weapon appears in hand after a
noticeable and seemingly non-constant delay.

neon-dev: > However, the changes in this PR affect attacker-side movement modifiers only.

Yes, I did not notice this when first checking the changes. However, we should make it clear by renaming `shouldApplyMovementModifier` to something more specific. Especially since we will likely need different methods for future fixes on the defensive side.

> without access to client source code it is
difficult to determine the exact timing logic behind these state changes

I don't think knowing the client logic would give us reliable info here. The server does the damage calculation and there is no connection between the displayed modifier on the client side and the actual applied modifier on the server side.  
Our current implementation is a 1 second delay after receiving the last movement packet (initial delay is not implemented yet). The retail server also can only act on such network packets. The only question is what exact delay(s) it is using.

Birthright50: > Yes, I did not notice this when first checking the changes. However, we should make it clear by renaming `shouldApplyMovementModifier` to something more specific. Especially since we will likely need different methods for future fixes on the defensive side.

Renamed to `shouldApplyAttackerMovementModifier` but maybe you can suggest a better option.

> I don't think knowing the client logic would give us reliable info here. The server does the damage calculation and there is no connection between the displayed modifier on the client side and the actual applied modifier on the server side.

Yes, agreed — the server can only operate based on received movement packets, and the client-side indicator itself does not drive server-side calculations.

What I meant is slightly different: retail likely synchronizes client-side visual feedback and server-side logic using the same timing assumptions. The direction arrow shown on the client is not authoritative, but it reflects when the client considers movement state to be active.

While this does not give us exact server logic, it can still provide a rough reference point for timing. I tried to figure out frame by frame when this arrow appears - ~800-850 ms after the start of movement

This does not prove the exact server delay, but this is the maximum we have.

neon-dev: > Renamed to `shouldApplyAttackerMovementModifier` but maybe you can suggest a better option.

I can't think of a better name right now, so let's got with it.

> While this does not give us exact server logic, it can still provide a rough reference point for timing. I tried to figure out frame by frame when this arrow appears - ~800-850 ms after the start of movement

This is a good starting point and we could use it for an initial implementation of the start delay. Feel free to create an issue or PR about it and thanks for taking the time to contribute👍

neon-dev:
I've rechecked everything and your changes here were complete.
~A similar fix still needs to be implemented for excluding NoReduceSpellATKInstantEffects from being affected by the target's movement modifiers, but this should be a separate issue or PR.~

neon-dev:
Correction about my last comment regarding NoReduceSpellATKInstantEffects: Further tests showed that they must be excluded from all reductions or boosts, which has been resolved with 8060313.

PR #99, opened 2025-12-21
Fix revision distance rule apply: only for players

Birthright50:
### Summary
Limit revision distance mechanics to player targets only.

Revision distance is no longer applied when casting spells on NPC targets.

### Additional Notes
Revision distance is now applied to player targets and player-owned summons.

It is currently unclear whether this mechanic should also apply to summons
created by NPCs. No change was made for NPC-owned summons until this behavior
is confirmed.


neon-dev:
SummonedObjects are spawns like Holy Servant. I've changed the check to cover anything controlled by a player. I also didn't check summons created by NPCs, since they're just regular NPCs which never seem to be affected by the revision distance.

Issue #11, opened 2024-11-27
Geo shield warning when starting game server


firestarter-77:
Everytime when i loading "GameServer", in console i see this geodata error.
![7771](https://github.com/user-attachments/assets/373c1790-65e3-474c-8faf-d0ab7a3e7b83)
Reloading files the same problem.

neon-dev:
This is a known warning about shields that can't be automatically matched to a siege location. This means that the siege status does not affect the shield status and is limited to a few (non-fortress) shields, like the platforms above the Abyss core.
We'll not look into it any time soon as it is a minor issue, but pull requests are always welcome.

firestarter-77:
Ok, same problem in 3 files: 210050000.geo, 310100000.geo, 400010000.geo.

SVDNESS:
Part 1.
You need to make the logic in the "attachShield" method less fragile:
1. Check all zones, not just getFirst().
2. Check corners, not just the center.
Because:
your original binding of "shield -> siege location" in the core is made according to a very strict and sometimes incorrect condition: "the center of the shield must fall inside the first zone."
1. getFirst() is not a guarantee of the "main" zone.
2. Checking only the center — it often misses.
Why do "AABB angles" decide?
If at least one corner of AABB is inside the zone, then the object intersects the zone (or at least its bounding box intersects).
And for the purposes of "attaching a shield to a location," this is much closer to the truth than "the center must be inside."
This is important because siegeLocationId is the key identifier, which determines whether the shield is turned on|off and the logic of collisions (via observer).
When the shield remains LocId=0, then the "ghost" problems begin.

Part 2.
You also need to exclude maps that report a "WARN" error in the console.
But this is not a mistake!
These geo-shields should NOT be tied to SiegeLocation.
In any case, it was not worth trying to fix them through attachShield!
Because it would be logically wrong:
1. They don't have a SiegeLocation.
2. Assign them siegeLocationId -> break the logic.
Therefore, they just had to be eliminated.

It was also worth adding some kind of insurance, in the form of "public void clearObservers() {...}", this does not break the logic, but makes it more secure.
I am attaching 2 files to you, how it should look right in your project logic.

[ShieldService.java](https://github.com/user-attachments/files/24339752/ShieldService.java)
[SiegeShield.java](https://github.com/user-attachments/files/24339751/SiegeShield.java)

PR #104, opened 2025-12-26
Rework Siege Shield Logic

SVDNESS:
Attempt to solve #11

Part 1.
You need to make the logic in the "attachShield" method less fragile:

1. Check all zones, not just getFirst().
2. Check corners, not just the center.
Because:
your original binding of "shield -> siege location" in the core is made according to a very strict and sometimes incorrect condition: "the center of the shield must fall inside the first zone."
3. getFirst() is not a guarantee of the "main" zone.
4. Checking only the center — it often misses.
Why do "AABB angles" decide?
If at least one corner of AABB is inside the zone, then the object intersects the zone (or at least its bounding box intersects).
And for the purposes of "attaching a shield to a location," this is much closer to the truth than "the center must be inside."
This is important because siegeLocationId is the key identifier, which determines whether the shield is turned on|off and the logic of collisions (via observer).
When the shield remains LocId=0, then the "ghost" problems begin.

Part 2.
You also need to exclude maps that report a "WARN" error in the console.
But this is not a mistake!
These geo-shields should NOT be tied to SiegeLocation.
In any case, it was not worth trying to fix them through attachShield!
Because it would be logically wrong:

1. They don't have a SiegeLocation.
2. Assign them siegeLocationId -> break the logic.

Therefore, they just had to be eliminated.
It was also worth adding some kind of insurance, in the form of "public void clearObservers() {...}", this does not break the logic, but makes it more secure.

### DIFF
--- a/game-server/src/com/aionemu/gameserver/services/ShieldService.java
+++ b/game-server/src/com/aionemu/gameserver/services/ShieldService.java
@@ -1,72 +1,54 @@
 package com.aionemu.gameserver.services;
 
-import com.aionemu.gameserver.configs.main.GeoDataConfig;
-import com.aionemu.gameserver.controllers.observer.ActionObserver;
-import com.aionemu.gameserver.controllers.observer.CollisionDieActor;
-import com.aionemu.gameserver.controllers.observer.ShieldObserver;
-import com.aionemu.gameserver.dataholders.DataManager;
-import com.aionemu.gameserver.model.gameobjects.Creature;
-import com.aionemu.gameserver.model.siege.FortressLocation;
-import com.aionemu.gameserver.model.siege.SiegeLocation;
-import com.aionemu.gameserver.model.siege.SiegeShield;
-import com.aionemu.gameserver.model.siege.SiegeType;
-import com.aionemu.gameserver.model.templates.shield.ShieldTemplate;
-import com.aionemu.gameserver.geoEngine.bounding.BoundingBox;
-import com.aionemu.gameserver.geoEngine.bounding.BoundingVolume;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 
-/**
- * @author SVDNESS
- * @version 4.8 [JDK 25]
- */
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
+import com.aionemu.gameserver.configs.main.GeoDataConfig;
+import com.aionemu.gameserver.controllers.observer.ActionObserver;
+import com.aionemu.gameserver.controllers.observer.CollisionDieActor;
+import com.aionemu.gameserver.controllers.observer.ShieldObserver;
+import com.aionemu.gameserver.dataholders.DataManager;
+import com.aionemu.gameserver.geoEngine.bounding.BoundingBox;
+import com.aionemu.gameserver.model.gameobjects.Creature;
+import com.aionemu.gameserver.model.siege.FortressLocation;
+import com.aionemu.gameserver.model.siege.SiegeLocation;
+import com.aionemu.gameserver.model.siege.SiegeShield;
+import com.aionemu.gameserver.model.templates.shield.ShieldTemplate;
+
+/**
+ * @author xavier, Rolandas, SVDNESS
+ */
 public class ShieldService {
-	Logger log = LoggerFactory.getLogger(ShieldService.class);
+
+	private static final Logger log = LoggerFactory.getLogger(ShieldService.class);
+	private static final Map<Integer, Set<String>> IGNORED_SHIELDS_BY_MAP_ID = Map.of(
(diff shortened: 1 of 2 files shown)


game-server/src/com/aionemu/gameserver/services/ShieldService.java
```diff
@@ -1,108 +1,175 @@
...
-		protected static final ShieldService instance = new ShieldService();
-	}
-
 	private final Map<Integer, ShieldTemplate> sphereShields = new ConcurrentHashMap<>();
 	private final Map<Integer, List<SiegeShield>> registeredShields = new ConcurrentHashMap<>();
-
-	public static ShieldService getInstance() {
-		return SingletonHolder.instance;
-	}
+	private static final Set<Integer> IGNORE_DETACHED_SHIELDS_MAPS = Set.of(400010000, 310100000);
```
neon-dev: Whole maps should never be ignored, as this will hide any problems with detached shields on these maps.

game-server/src/com/aionemu/gameserver/services/ShieldService.java
```diff
@@ -1,108 +1,175 @@
...
+	private static final String[] IGNORE_DETACHED_SHIELDS_PREFIXES = {
+            "PR_A_AIRBUNKER_EFFECT_",
```
neon-dev: `PR_A_AIRBUNKER_EFFECT_` must not be ignored, as these belong to fortresses like Asteria or Divine Fortress.

game-server/src/com/aionemu/gameserver/services/ShieldService.java
```diff
@@ -1,108 +1,175 @@
...
+            "BU_AB_SAMJUNG_BASE_01_SHIELD_",
+            "BU_DG_DRAGONSHIELD_"
```
neon-dev: `BU_DG_DRAGONSHIELD_` does not have to be ignored, as it isn't logged anyway.

neon-dev (changes requested):
I have a few remarks (in addition to my code comments) that should be addressed first, before I can completely review this PR:

1. Please reformat the code and import order, so we can maintain a consistent code style. With IntelliJ, you can simply autoformat the files. With other IDEs, you would first need to make sure that [Editorconfig](https://editorconfig.org/#pre-installed) support is active.
2. Add your name to the list of authors but don't remove others if any of the old code remains (even in a refactored state).
3. Don't add the version to files, as it doesn't add useful information.
4. Don't use the `var` keyword too exessively, as hiding the variable type can hurt maintainability.
5. Please remove the null checks you added for non-nullable objects (like for getGeometry, for example). Checking such objects for null would hide problems if they ever were null. We can discuss adding `Objects.requireNonNull` in certain constructors but that is out of scope for this pull request.

Thank you

SVDNESS:
Of course! I just showed and wrote "how it could look in your project."
In the final form, it would be something like this:
    private static final Map<Integer, Set<String>>
IGNORE_DETACHED_SHIELDS_EXACT = Map.of(
            210050000, Set.of("BU_DG_DRGONCASTLESHIELD_04D_220732"),
            400010000, Set.of(
                    "BU_AB_SAMJUNG_BASE_01_SHIELD_313626",
                    "BU_AB_SAMJUNG_BASE_01_SHIELD_299314",
                    "BU_AB_SAMJUNG_BASE_01_SHIELD_137227",
                    "PR_A_AIRBUNKER_EFFECT_01A_CHILD1_159556",
                    "PR_A_AIRBUNKER_EFFECT_01A_CHILD2_159556",
                    "PR_A_AIRBUNKER_EFFECT_01A_CHILD1_324011",
                    "PR_A_AIRBUNKER_EFFECT_01A_CHILD2_324011",
                    "PR_A_AIRBUNKER_EFFECT_01A_CHILD1_312167",
                    "PR_A_AIRBUNKER_EFFECT_01A_CHILD2_312167"
            ),
            310100000, Set.of("BU_AB_CASTLESHIELD_SAMJUNG_03C_TYPE2_487543")
    );

private static boolean shouldIgnoreDetachedShield(int mapId, SiegeShield
shield) {
        var geo = shield.getGeometry();
        if (geo == null) {
            return false;
        }
        var name = geo.getName();
        if (name == null) {
            return false;
        }
        var exact = IGNORE_DETACHED_SHIELDS_EXACT.get(mapId);
        return exact != null && exact.contains(name);
    }

сб, 27 дек. 2025 г. в 07:33, Neon ***@***.***>:

SVDNESS:
Unfortunately, I don't have time for corrections right now. A lot of work
besides AION. If you have time and my edits are useful to you, you can just
add them to yourself. And don't include me in the authors. Or specify a
co-author.
I'm not chasing fame or anything else)

сб, 27 дек. 2025 г. в 10:11, SVDNESS ***@***.***>:

neon-dev:
I have reverted the SiegeShield changes because players must always be observed, even when the fortress is not under shield yet.
Otherwise players in the fortress area will not be affected by the newly activated shields.

I've also resolved most of my earlier comments and made it so that ignored shields are no longer added.
The only remaining problem is with airbunkers which can not be attached because they are too far away from the fortress zone.

neon-dev:
I went ahead and merged this PR ins its current form, as it is a great improvement. We'll have to investigate the airbunker issue separately.

PR #107, opened 2025-12-26
Prevent misconfiguration of NIO threads > 1

QTimort:
## Summary
- Fail startup if `gameserver.network.nio.threads` > 1 (game server is not thread-safe)
- Added bypass config `gameserver.network.nio.threads.unsafe.allow` for dev/debugging

Let's stop the delusion.

### DIFF
--- a/game-server/src/com/aionemu/gameserver/configs/network/NetworkConfig.java
+++ b/game-server/src/com/aionemu/gameserver/configs/network/NetworkConfig.java
@@ -68,6 +68,12 @@ public class NetworkConfig {
 	public static int NIO_READ_WRITE_THREADS;
 
 	/**
+	 * Allows values > 1 for nio.threads. Only for development/debugging - the game server is not thread-safe!
+	 */
+	@Property(key = "gameserver.network.nio.threads.unsafe.allow", defaultValue = "false")
+	public static boolean NIO_READ_WRITE_THREADS_UNSAFE_ALLOW;
+
+	/**
 	 * Number of minimum threads that will be used to execute aion client packets.
 	 */
 	@Property(key = "gameserver.network.packet.processor.threads.min", defaultValue = "4")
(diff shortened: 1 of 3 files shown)


game-server/src/com/aionemu/gameserver/configs/network/NetworkConfig.java
```diff
@@ -67,6 +67,12 @@ public class NetworkConfig {
 	@Property(key = "gameserver.network.nio.threads", defaultValue = "1")
 	public static int NIO_READ_WRITE_THREADS;
 
+	/**
+	 * Allows values > 1 for nio.threads. Only for development/debugging - the game server is not thread-safe!
+	 */
+	@Property(key = "gameserver.network.nio.threads.unsafe.allow", defaultValue = "false")
```
neon-dev, line 73: This would need to be added to network.properties as well, but I'm thinking it would be even better to remove the `gameserver.network.nio.threads` altogether until it's fixed.
It's only useful to developers to set invalid values and they can do this in code.

neon-dev (changes requested):

game-server/src/com/aionemu/gameserver/configs/network/NetworkConfig.java
QTimort: The idea was to keep it somewhat backward compatible and still configurable without having to change the code and recompile. it's just to be very clear and explicit that setting values higher than 1 shouldn't be used in prod

neon-dev: I don't have a strong preference for either solution.
If you think it's beneficial to add the new config, then add it to network.properties as well. Otherwise, remove both and hardcode the value.

QTimort:
@neon-dev should now be mergeable based on your prev comment

PR #113, opened 2025-12-29
Fix delevel exploit while using exp extracting items

daddycaddy:
prevents delevel exploit (only extract exp progress inside current level) and only if the player has the correct amount of exp


neon-dev:
I've changed it so that `act()` uses the same checks as `canAct()` and added a separate check that the extractor item has been consumed, which makes everything safe against timing based exploits.
The system messages are now also the correct ones.

PR #116, opened 2026-01-02
Plume Fix

daddycaddy:
Not disappearing when tempering fails and plume is equipped


neon-dev:
I introduced this bug with aa9c307 and didn't notice, as destroyed equipped items only reappeared in the inventory after relogging.
This has been corrected and so your fix that only worked for plumes is no longer needed.

PR #121, opened 2026-01-12
Fix magical cast time calculation: proper stat cap (75%) and summon-type behavior

Birthright50:
## Description

This PR reworks the magical cast time calculation to better match retail Aion behavior.

### Key changes

- **Casting speed stats are hard-capped at 75% reduction**
  Stat-based cast speed (e.g. gear) is capped before any other modifiers are applied, matching retail behavior.

- **Skill-based cast speed buffs are applied additively**
  Buffs (`BOOST_CASTING_TIME_SKILL` and subtype-specific bonuses) do not modify stat values and are applied on top of the capped stat reduction.

- **Different final caps based on skill subtype**
  - **Normal skills**: final cast speed reduction capped at **75%** (minimum 25% of base cast time)
  - **SUMMON / SUMMONHOMING / SUMMONTRAP**: can exceed 75% and reach **100% reduction (instant cast)**

- **Removed hardcoded abnormal-effect checks**
  Previous logic relied on specific effect IDs to bypass the cast time cap. Retail behavior shows that summon-type skills can reach 100% reduction without such effects.

- **Corrected base cast duration floor**
  The lower bound was corrected from `30%` to the correct `25%` of base cast duration.

- **Simplified and clarified calculation flow**
  The logic now follows a clear order:
  1. Apply stat-based cap
  2. Apply skill-based buffs
  3. Apply final cap based on skill subtype

  Legacy workaround code related to trap skills was removed.

### Examples (retail behavior)

- **Gear −30% + buff −50%**
  - Normal skill: **75% cap**
  - Summon skill: **80% reduction**

- **Gear −80% + buff −50%**
  - Normal skill: **75% cap**
  - Summon skill: **100% reduction**

- **No gear + buff −100%**
  - Normal skill: **75% cap**
  - Summon skill: **100% reduction**
  
- **Gear −80% + buff −5%**
  - Stat reduction is capped to **75%** first
  - Total reduction becomes **80%**
  - Normal skill: **75% cap**
  - Summon skill: **80% reduction**

### Note
The existing workaround for `SUMMONTRAP` cast speed calculation was intentionally left unchanged.
It may be redundant with the updated logic, but was preserved due to unclear historical context.

[Video from 4.6 PTS](https://www.youtube.com/watch?v=Wnl25YkPOuQ)

### DIFF
--- a/game-server/src/com/aionemu/gameserver/skillengine/model/Skill.java
+++ b/game-server/src/com/aionemu/gameserver/skillengine/model/Skill.java
@@ -349,7 +349,7 @@ public class Skill {
 			castDuration = (int) effector.getGameStats().getPositiveStat(StatEnum.ATTACK_SPEED, baseCastDuration);
 		else
 			castDuration = calculateMagicalCastDuration();
-		return Math.max(castDuration, (int) (baseCastDuration * 0.3f)); // TODO check limit with retail
+		return Math.max(castDuration, (int) (baseCastDuration * 0.25f));
 	}
 
 	private int calculateCastDuration() {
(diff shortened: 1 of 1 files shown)


neon-dev:
> The existing workaround for SUMMONTRAP cast speed calculation was intentionally left unchanged.
> It may be redundant with the updated logic, but was preserved due to unclear historical context.

I have verified that this is no longer needed and removed it to prevent potential calculation errors (depending on the player's buffs).

PR #123, opened 2026-01-17
Fixed max speed chair bug when standing

crlfrancis:


game-server/src/com/aionemu/gameserver/network/aion/clientpackets/CM_EMOTION.java
```diff
@@ -157,8 +157,10 @@ protected void runImpl() {
 				player.setState(CreatureState.CHAIR, true);
 				break;
 			case CHAIR_UP:
-				if (player.isInState(CreatureState.CHAIR))
+				if (player.isInState(CreatureState.CHAIR)) {
+					player.unsetState(CreatureState.CHAIR);
 					player.setState(CreatureState.ACTIVE);
+				}
```
neon-dev: Let's replace the pseudo state `CHAIR` with `ACTIVE` as I originally intended in f2ad6d7:
```suggestion
				if (player.isInState(CreatureState.CHAIR))
					player.setState(CreatureState.ACTIVE, true);
```

PR #125, opened 2026-01-19
Fix PvP/PvE ratio caps and damage multiplier handling

Birthright50:
This PR fixes PvP/PvE damage ratio handling by applying caps to aggregated attack and defense ratios instead of individual stats.

On retail, PvP/PvE attack and defense ratios (base + physical/magical modifiers) are capped only after aggregation.
Using per-stat lower/upper cap helpers would incorrectly clamp individual components and break valid mixed-stat configurations.

The final damage multiplier is additionally clamped to prevent negative or zero damage values, while upper bounds remain governed by stat caps.

lowerCapFor / upperCapFor functions operate on individual stats,
while PvP/PvE ratio limits on retail apply to the final aggregated value
(base + physical/magical modifiers).
Using per-stat caps would change valid configurations and result in incorrect behavior.

PvP Caps
```
PVP_ATTACK_RATIO + PVP_ATTACK_RATIO_PHYSICAL | PVP_ATTACK_RATIO_MAGICAL
∈ [-900 … 1000]   // -90% … +100%

PVP_DEFEND_RATIO + PVP_DEFEND_RATIO_PHYSICAL | PVP_DEFEND_RATIO_MAGICAL
∈ [-1000 … 900]   // -100% … +90%
```

PvE caps: 
```
PVE_ATTACK_RATIO + PVE_ATTACK_RATIO_PHYSICAL | PVE_ATTACK_RATIO_MAGICAL
∈ [-900 … 5000]   // -90% … +500%

PVE_DEFEND_RATIO+ PVE_DEFEND_RATIO_PHYSICAL | PVE_DEFEND_RATIO_MAGICAL
∈ [-5000 … 900]   // -500% … +90%
```
Final Damage Multiplier:
```
multiplier = 1 + (attackRatio - defenseRatio) / 1000
multiplier >= 0.1
```
For example:
[PvEDamageRatio + PvEDamageRatio_magical cap = 5000
](https://www.youtube.com/watch?v=IsLj05BrOus)

[PvPDamageRation min and max cap](https://www.youtube.com/watch?v=Kh2krfFXo2g)


game-server/src/com/aionemu/gameserver/model/stats/calc/StatCapUtil.java
```diff
@@ -57,6 +59,26 @@ private static void calculate(Stat2 stat2, int lowerCap, int upperCap) {
...
+				case ATTACK -> new Cap(-900, 1000);
+				case DEFENSE -> new Cap(-1000, 900);
+			};
+			case PVE -> switch (type) {
+				case ATTACK -> new Cap(-900, 5000);
+				case DEFENSE -> new Cap(-5000, 900);
+			};
+		};
+
+		return Math.min(cap.max(), Math.max(cap.min(), value));
```
neon-dev:
```suggestion
		return Math.clamp(value, cap.min(), cap.max());
```

game-server/src/com/aionemu/gameserver/utils/stats/StatFunctions.java
```diff
@@ -445,46 +443,54 @@ public static int getApNpcRating(NpcRating npcRating) {
...
+					defenseBonus += target.getGameStats().getStat(StatEnum.PVE_DEFEND_RATIO_MAGICAL, 0).getCurrent();
 			}
 		}
-		return damage + (damage * attackBonus) - (damage * defenseBonus);
+		CombatMode mode = pvpTarget ? CombatMode.PVP : CombatMode.PVE;
+		// PvP/PvE ratio caps are applied after aggregation (retail behavior)
+		attackBonus = StatCapUtil.limitValueForPvpOrPveStat(mode, RatioType.ATTACK, attackBonus);
+		defenseBonus = StatCapUtil.limitValueForPvpOrPveStat(mode, RatioType.DEFENSE, defenseBonus);
+		float multiplier = 1f + (attackBonus - defenseBonus) / 1000f;
+		// prevent negative or zero damage
```
neon-dev: This comment should be rephrased to make it clear that this is retail behavior and the value of 10% is not arbitrary.

neon-dev (changes requested):

PR #126, opened 2026-01-21
NPC diff lvl modifiers have been adjusted

Birthright50:
When difflvl with npc > 11 damage reduced to 1 (if no shields)
Proof: Player 56 lvl, training dummy 68. (For SpellATK, Bleed, MagicCounterATK, Poison shouldn't work)
image


neon-dev:
I've auto-formatted your changes, as the current formatting rules don't allow for single-line ifs.

The new values have been validated and are correct, but I also noticed that it seems like this modifier should be applied to the calculated attack value instead of the final damage.
It certainly needs further tests, but my limited testing showed that the effective (physical) damage reduction percentage increased by more than 10% per level difference, if the target had (physical) defense > 0.

PR #128, opened 2026-01-27
Unify elemental resistance caps and damage normalization logic

Birthright50:
### Summary
Refactored elemental resistance handling to centralize caps, remove magic numbers, and prevent resistance values from exceeding valid bounds after situational modifiers.

### Key changes
- Centralized elemental resistance base value and level-based scaling logic.
- Unified elemental resistance caps for players and NPCs via `StatCapUtil`.
- Added explicit stat clamping after movement/situational modifiers in combat formulas.
- Moved elemental resistance normalization (denominator) logic into combat context.
- Improved method naming and JavaDoc to clearly reflect game mechanics.

### Behavior
- NPCs always use a fixed elemental resistance scale (1300).
- Player elemental resistance caps and damage normalization scale with level above 50.
- Elemental resistance is clamped both at stat calculation time and after combat modifiers.

For example, NPC can reduce all damage if he has 1300 elemental resistance, or he will be dealt double damage if he has -1300.

### DIFF
--- a/game-server/data/handlers/admincommands/Info.java
+++ b/game-server/data/handlers/admincommands/Info.java
@@ -100,12 +100,12 @@ public class Info extends AdminCommand {
 							+ "\n\t\tEvasion: " + pgs.getEvasion().getCurrent()
 							+ "\n\t\tCrit. Strike Resist: " + pgs.getPCR().getCurrent()
 							+ "\n\t\tCrit. Strike Fortitude: " + pgs.getStat(StatEnum.PHYSICAL_CRITICAL_DAMAGE_REDUCE, 0).getCurrent()
-							+ "\n\t\tWind Resist: " + pgs.getElementalDefendFor(SkillElement.WIND)
-							+ "\n\t\tWater Resist: " + pgs.getElementalDefendFor(SkillElement.WATER)
-							+ "\n\t\tEarth Resist: " + pgs.getElementalDefendFor(SkillElement.EARTH)
-							+ "\n\t\tFire Resist: " + pgs.getElementalDefendFor(SkillElement.FIRE)
-							+ "\n\t\tDark Resist: " + pgs.getElementalDefendFor(SkillElement.DARK)
-							+ "\n\t\tLight Resist: " + pgs.getElementalDefendFor(SkillElement.LIGHT)
+							+ "\n\t\tWind Resist: " + pgs.getElementalDefenseFor(SkillElement.WIND)
+							+ "\n\t\tWater Resist: " + pgs.getElementalDefenseFor(SkillElement.WATER)
+							+ "\n\t\tEarth Resist: " + pgs.getElementalDefenseFor(SkillElement.EARTH)
+							+ "\n\t\tFire Resist: " + pgs.getElementalDefenseFor(SkillElement.FIRE)
+							+ "\n\t\tDark Resist: " + pgs.getElementalDefenseFor(SkillElement.DARK)
+							+ "\n\t\tLight Resist: " + pgs.getElementalDefenseFor(SkillElement.LIGHT)
 							+ "\n\t-------------PvP Stats-------------"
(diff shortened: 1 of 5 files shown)


game-server/src/com/aionemu/gameserver/model/stats/calc/StatCapUtil.java
```diff
@@ -90,13 +120,8 @@ private static int lowerCapFor(StatEnum stat) {
...
 					break;
-				case WATER_RESISTANCE:
-				case FIRE_RESISTANCE:
-				case EARTH_RESISTANCE:
-				case WIND_RESISTANCE:
-				case DARK_RESISTANCE:
-				case LIGHT_RESISTANCE:
-					value = -1150;
+				case HEAL_BOOST:
+					value = -1000;
```
neon-dev: This new `HEAL_BOOST` min cap seems unrelated. Is it maybe supposed to be part of a different PR?

game-server/data/handlers/admincommands/Info.java
```diff
@@ -100,12 +100,12 @@ public void execute(Player admin, String... params) {
 							+ "\n\t\tEvasion: " + pgs.getEvasion().getCurrent()
 							+ "\n\t\tCrit. Strike Resist: " + pgs.getPCR().getCurrent()
 							+ "\n\t\tCrit. Strike Fortitude: " + pgs.getStat(StatEnum.PHYSICAL_CRITICAL_DAMAGE_REDUCE, 0).getCurrent()
-							+ "\n\t\tWind Resist: " + pgs.getMagicalDefenseFor(SkillElement.WIND)
-							+ "\n\t\tWater Resist: " + pgs.getMagicalDefenseFor(SkillElement.WATER)
-							+ "\n\t\tEarth Resist: " + pgs.getMagicalDefenseFor(SkillElement.EARTH)
-							+ "\n\t\tFire Resist: " + pgs.getMagicalDefenseFor(SkillElement.FIRE)
-							+ "\n\t\tDark Resist: " + pgs.getMagicalDefenseFor(SkillElement.DARK)
-							+ "\n\t\tLight Resist: " + pgs.getMagicalDefenseFor(SkillElement.LIGHT)
+							+ "\n\t\tWind Resist: " + pgs.getElementalDefendFor(SkillElement.WIND)
```
neon-dev: Please rename all occurrences of `ElementalDefend` to either `ElementalDefense` or `ElementalResistance`. Besides the grammar, this is also more consistent with similar methods that handle defensive stats.

neon-dev (changes requested):

game-server/data/handlers/admincommands/Info.java
Birthright50: These names are based on retail terminology (e.g. ElementalDefendFire), so I kept them for consistency with the original data.

However, I agree that "Defense" or "Resistance" would be more idiomatic in Java.

game-server/src/com/aionemu/gameserver/utils/stats/StatFunctions.java
```diff
@@ -336,24 +336,49 @@ public static List<AttackResult> calculateAttackDamage(Creature attacker,
...
+		float elementalDenominator = getElementalDefenseDenominator(effector, attacked);
+		int rawDefense = attacked.getGameStats().getElementalDefenseFor(element);
+		int adjustedDefense = (int) adjustStatByMovementModifier(attacked, element.getStatForElement(), rawDefense);
+		adjustedDefense = StatCapUtil.clampStatValue(element.getStatForElement(), attacked, adjustedDefense);
+		return damage * (1f - adjustedDefense / elementalDenominator);
+	}
+
+	private static int getElementalDefenseDenominator(Creature effector, Creature attacked) {
+		if (attacked instanceof Player) {
+			int level = Math.min(effector.getLevel(), attacked.getLevel());
```
neon-dev, line 378: This `Math.min` is a pretty odd mechanic as it reduces the received elemental damage even further if the target has a lower level than the attacker (and the elemental resistance is below the upper bound), but I verified it on 4.6 PTS.

PR #131, opened 2026-03-04
Fears and slow effects

jvbhidalgo:
fear removes dot that reduces mov speed, and while on fear dots that causes slow only do the first damage, the skill dont stick to the enemy. Validation for fears that reduces move speed only (curse of fire/water, shriek etc.)

evidence: https://youtu.be/EyhOOToSReU?si=D4LSKofgpD2RIPFp&t=243 ik its on 4.6 but i was told that this was on 4.8 too

### DIFF
--- a/game-server/src/com/aionemu/gameserver/controllers/effect/EffectController.java
+++ b/game-server/src/com/aionemu/gameserver/controllers/effect/EffectController.java
@@ -94,6 +94,9 @@ public class EffectController {
 		}
 		put(mapToUpdate, nextEffect);
 
+		if (nextEffect.isFearWithSpeedReduction())
+			removeSpeedReductionEffects();
+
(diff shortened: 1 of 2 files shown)


neon-dev:
Hey, sorry for the long wait.
I've validated the behavior on 4.6 now, but in 4.6 only slows of type Snare are removed. Slows of type StatDown, for example, are not removed by the fear skill.
It also worked like this in the emulator up until version 4.7.5, because the fear skills themselves applied slows of type Snare with the same `effectid`. So these were regular effect overrides.
In 4.8, fear skills got changed to use slows of type StatDown instead of Snare. They also have a different `effectid` now, which is why they no longer remove other slow effects.
For that reason, I am fairly confident that the current behavior is correct.

PR #134, opened 2026-05-03
Mvn wrapper

Birthright50:
Add Maven Wrapper


neon-dev:
Unfortunately, this is not something we're interested in adding to our repo.
Developers who do not use a Java IDE will have to set up Maven externally.

Birthright50:
> Unfortunately, this is not something we're interested in adding to our repo. Developers who do not use a Java IDE will have to set up Maven externally.

Wrapper ensures consistent Maven version across all environments, not just for developers without IDE.

neon-dev:
It is not guaranteed that the maven wrapper is automatically used by all IDEs (or developers). Eclipse doesn't use it by default and even with IntelliJ it depends on how the project is opened.
If a consistent Maven version was a concern, I think something like the `maven-enforcer-plugin` would be a better alternative. 
I just tested building this project with older versions and the `git-commit-id-maven-plugin` requires a Maven version >= 3.6.3, which is almost seven years old. Maven 3.6.3 builds this project just fine, so I think we're good.

Birthright50:
> It is not guaranteed that the maven wrapper is automatically used by all IDEs (or developers). Eclipse doesn't use it by default and even with IntelliJ it depends on how the project is opened.

Wrapper is not intended to be automatically picked up by IDEs, but to provide a consistent CLI/CI build environment.

> If a consistent Maven version was a concern, I think something like the `maven-enforcer-plugin` would be a better alternative.

`maven-enforcer-plugin` only validates the Maven version, but still requires Maven to be installed manually. Wrapper ensures both the correct version and that it is available without additional setup.

> I just tested building this project with older versions and the `git-commit-id-maven-plugin` requires a Maven version >= 3.6.3, which is almost seven years old. Maven 3.6.3 builds this project just fine, so I think we're good.

Minimum required Maven version of a single plugin does not guarantee overall build reproducibility or compatibility across environments. While it works with older Maven versions today, the wrapper ensures future consistency and avoids unexpected issues caused by environment differences.

And wrapper removes the need for developers to install Maven manually. I personally don’t install Maven globally and rely on the wrapper, as it simplifies setup.

neon-dev:
I mostly agree and understand where you're coming from, but let me explain.

Both Java IDEs and Java CI environments typically include Maven. And while CLI users may be used to using `./mvnw`, this is not the case for everyone and certainly not for most IDE users. CI also only uses it if it is configured to do so.
Sure, we could change commands in the workflow scripts and the README.md to make a wrapper the default entrypoint, but it doesn't seem worth it at the moment.
For these reasons, I view Maven wrappers as more of a nice-to-have for some users, rather than a must-have in a project.

Besides, Maven 3 versions are known to have very good backwards compatibility, so it is currently not important if an older minor version is used.
If build reproducibility or compatibility becomes an issue, I think a combination of all mentioned steps (including adding the wrapper and the enforcer plugin) would be the best solution.

I hope you can accept this reasoning, even if you disagree.

PR #135, opened 2026-05-03
fix: align cumulative resist logic with retail

Birthright50:
## Summary
Fix cumulative resist calculation and align with retail formulas.

## Formulas

baseDuration = remain - rnd(0, randomTime)  
cooldown = baseDuration * multiplier + flat  
finalDuration = baseDuration * durationMultiplier * pvpRemain

Retail **repeated_abnormal_status_immune.xml**:
<details>
<summary>Retail XML reference</summary>

```xml
<repeated_abnormal_status_immunes>
	<repeated_abnormal_status_immune>
		<id>2</id>
		<name>PARALYZE</name>
		<holding_time1>1</holding_time1>
		<holding_time2>0</holding_time2>
		<resist_value1>0</resist_value1>
		<resist_value2>0</resist_value2>
		<resist_value3>200</resist_value3>
		<resist_value4>400</resist_value4>
		<resist_value5>1000</resist_value5>
		<time_value1>100</time_value1>
		<time_value2>90</time_value2>
		<time_value3>85</time_value3>
		<time_value4>80</time_value4>
		<time_value5>0</time_value5>
	</repeated_abnormal_status_immune>
	<repeated_abnormal_status_immune>
		<id>3</id>
		<name>SLEEP</name>
		<holding_time1>1</holding_time1>
		<holding_time2>0</holding_time2>
		<resist_value1>0</resist_value1>
		<resist_value2>0</resist_value2>
		<resist_value3>200</resist_value3>
		<resist_value4>400</resist_value4>
		<resist_value5>1000</resist_value5>
		<time_value1>100</time_value1>
		<time_value2>90</time_value2>
		<time_value3>85</time_value3>
		<time_value4>80</time_value4>
		<time_value5>0</time_value5>
	</repeated_abnormal_status_immune>
	<repeated_abnormal_status_immune>
		<id>9</id>
		<name>FEAR</name>
		<holding_time1>1</holding_time1>
		<holding_time2>2000</holding_time2>
		<resist_value1>0</resist_value1>
		<resist_value2>0</resist_value2>
		<resist_value3>200</resist_value3>
		<resist_value4>400</resist_value4>
		<resist_value5>1000</resist_value5>
		<time_value1>100</time_value1>
		<time_value2>90</time_value2>
		<time_value3>85</time_value3>
		<time_value4>80</time_value4>
		<time_value5>0</time_value5>
	</repeated_abnormal_status_immune>
</repeated_abnormal_status_immunes>

</details>

### DIFF
--- a/game-server/src/com/aionemu/gameserver/configs/main/CustomConfig.java
+++ b/game-server/src/com/aionemu/gameserver/configs/main/CustomConfig.java
@@ -238,6 +238,12 @@ public class CustomConfig {
 	public static int CHARACTER_DELETION_TIME_MINUTES;
 
 	/**
+	 * Don't consume potions when already at full HP/MP
+	 */
+	@Property(key = "gameserver.items.ignore_potions_at_full_health", defaultValue = "false")
+	public static boolean IGNORE_POTIONS_AT_FULL_HEALTH;
+
+	/**
 	 * Custom Reward Packages
 	 */
 	@Property(key = "gameserver.custom.starter_kit.enable", defaultValue = "false")
--- a/game-server/src/com/aionemu/gameserver/configs/main/CustomConfig.java
+++ b/game-server/src/com/aionemu/gameserver/configs/main/CustomConfig.java
@@ -263,4 +263,10 @@ public class CustomConfig {
 
 	@Property(key = "gameserver.rates.godstone.evaluation.cooldown_millis", defaultValue = "750")
 	public static int GODSTONE_EVALUATION_COOLDOWN_MILLIS;
+
+	/**
+	 * PvP config
+	 */
+	@Property(key = "gameserver.pvp.cumulative_resist.count_summon_effects", defaultValue = "false")
+	public static boolean COUNT_SUMMON_EFFECTS_FOR_CUMULATIVE_RESIST;
 }
--- a/game-server/src/com/aionemu/gameserver/configs/main/CustomConfig.java
+++ b/game-server/src/com/aionemu/gameserver/configs/main/CustomConfig.java
@@ -269,4 +269,10 @@ public class CustomConfig {
 
 	@Property(key = "gameserver.rates.godstone.evaluation.cooldown_millis", defaultValue = "750")
 	public static int GODSTONE_EVALUATION_COOLDOWN_MILLIS;
+
+	/**
+	 * PvP config
+	 */
+	@Property(key = "gameserver.pvp.cumulative_resist.count_summon_effects", defaultValue = "false")
+	public static boolean COUNT_SUMMON_EFFECTS_FOR_CUMULATIVE_RESIST;
 }
--- a/game-server/src/com/aionemu/gameserver/configs/main/CustomConfig.java
+++ b/game-server/src/com/aionemu/gameserver/configs/main/CustomConfig.java
@@ -265,7 +265,7 @@ public class CustomConfig {
 	public static int GODSTONE_EVALUATION_COOLDOWN_MILLIS;
 
 	/**
-	 * PvP config
+	 * Count summon-applied abnormal effects for cumulative resist.
 	 */
 	@Property(key = "gameserver.pvp.cumulative_resist.count_summon_effects", defaultValue = "false")
 	public static boolean COUNT_SUMMON_EFFECTS_FOR_CUMULATIVE_RESIST;
--- a/game-server/src/com/aionemu/gameserver/controllers/effect/CumulativeResist.java
+++ b/game-server/src/com/aionemu/gameserver/controllers/effect/CumulativeResist.java
@@ -11,8 +11,9 @@ class CumulativeResist {
 			level++;
 		this.expirationTime = System.currentTimeMillis() + maxDurationMillis;
 	}
-	//time_value* from repeated_abnormal_status_immune.xml retail file
+
 	float getDurationMultiplier() {
+		//time_value* from repeated_abnormal_status_immune.xml retail file
 		return switch (level) {
 			case 0, 1 -> 1;
 			case 2 -> 0.9f;
@@ -21,16 +22,17 @@ class CumulativeResist {
 			default -> 0;
 		};
 	}
-
-	//holding_time2 from repeated_abnormal_status_immune.xml retail file
+	
 	int getCooldownTimeOffset(CumulativeResistType type) {
+		//holding_time2 from repeated_abnormal_status_immune.xml retail file
 		return switch (type) {
 			case SLEEP, PARALYZE -> 0;
 			case FEAR -> 2000;
 		};
 	}
-	//resist_value* from repeated_abnormal_status_immune.xml retail file
+
 	int getResistance() {
+		//resist_value* from repeated_abnormal_status_immune.xml retail file
 		resetIfExpired();
 		return switch (level) {
 			case 0, 1, 2 -> 0;
--- a/game-server/src/com/aionemu/gameserver/controllers/effect/CumulativeResist.java
+++ b/game-server/src/com/aionemu/gameserver/controllers/effect/CumulativeResist.java
@@ -22,7 +22,7 @@ class CumulativeResist {
 			default -> 0;
 		};
 	}
-	
+
 	int getCooldownTimeOffset(CumulativeResistType type) {
 		//holding_time2 from repeated_abnormal_status_immune.xml retail file
 		return switch (type) {
@@ -32,8 +32,8 @@ class CumulativeResist {
 	}
 
 	int getResistance() {
-		//resist_value* from repeated_abnormal_status_immune.xml retail file
 		resetIfExpired();
+		//resist_value* from repeated_abnormal_status_immune.xml retail file
 		return switch (level) {
 			case 0, 1, 2 -> 0;
 			case 3 -> 200;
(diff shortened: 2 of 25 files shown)

### CODE REVIEW

File: game-server/src/com/aionemu/gameserver/controllers/effect/CumulativeResist.java [hunk 1]
```diff
@@ -12,13 +12,24 @@ void tryIncrementLevel(long maxDurationMillis) {
 		this.expirationTime = System.currentTimeMillis() + maxDurationMillis;
 	}
 
-	float getDurationMultiplier() {
-		resetIfExpired();
+	float getDurationMultiplier(CumulativeResistType type) {
```
neon-dev:
```suggestion
	float getDurationMultiplier() {
```

game-server/src/com/aionemu/gameserver/controllers/effect/CumulativeResist.java
```diff
@@ -12,13 +12,24 @@ void tryIncrementLevel(long maxDurationMillis) {
...
+			case 2 -> 0.9f;
+			case 3 -> 0.85f;
+			case 4 -> 0.8f;
+			default -> 0;
+		};
+	}
+
+	int getCooldownMultiplier(CumulativeResistType type) {
+		return 1;
+	}
```
neon-dev: What's the purpose of this method?

game-server/src/com/aionemu/gameserver/controllers/effect/CumulativeResist.java
```diff
@@ -12,13 +12,24 @@ void tryIncrementLevel(long maxDurationMillis) {
...
 		return switch (level) {
-			case 0, 1, 2 -> 1f;
-			case 3 -> 0.9f;
-			case 4 -> 0.85f;
-			default -> 0.8f;
+			case 0, 1 -> 1;
+			case 2 -> 0.9f;
+			case 3 -> 0.85f;
+			case 4 -> 0.8f;
+			default -> 0;
```
neon-dev, line 22: What's causing the shift by -1 level here?
I remember having validated the level mapping for the previous implementation and don't see that this PR changes the increment logic, so it looks like it now applies a duration reduction starting from the second fear/para/sleep instead of the third.
Also, the fifth consecutive CC effect would now have a duration of zero, which I don't remember seeing when I was testing it last year.

Edit: I see now. The level shift does not affect the effective multipliers because this PR also changes `tryIncrementLevel()` to happen after calling `getDurationMultiplier()`. And the new default case should never be reached because on the fifth level, the spell will always be resisted.
In any case, these changes are a clear improvement in terms of clarity and consistency with the retail file.

neon-dev (changes requested):

game-server/src/com/aionemu/gameserver/controllers/effect/CumulativeResist.java
Birthright50: Overall, there is little point. this is just unification in accordance with the file, if in the future it will be possible to read values ​​from retail file, and not hardcode them in the code (although in essence there is little point in this).

game-server/src/com/aionemu/gameserver/controllers/effect/CumulativeResist.java
Birthright50: I edited the file, video with an example. https://www.youtube.com/watch?v=ObQ38iJlmnY

game-server/src/com/aionemu/gameserver/controllers/effect/CumulativeResist.java
neon-dev: Ah, I didn't know about the PTS file `repeated_abnormal_status_immune.xml` before seeing it today in the video of your other comment (I overlooked it in your PR description). So I assume this multiplier is the corresponding value for `<holding_time1>`, which is indeed always 1.
Better remove it until it is needed to keep the code a bit cleaner.

Birthright50: Right, `<holding_time1>` is the coefficient, `<holding_time2>` flat value in ms.

game-server/src/com/aionemu/gameserver/skillengine/model/Effect.java
```diff
@@ -845,14 +845,14 @@ private void stopPeriodicActions() {
 	private int calculateEffectsDuration() {
 		long duration = calculateTemplateDuration();
 
-		// adjust with pvp duration (not sure why some self target skills have pvp duration o.O idk how to handle that)
 		if (getEffected() instanceof Player effectedPlayer) {
-			if (skillTemplate.getPvpDuration() != 0 && !effector.equals(effected))
-				duration = duration * skillTemplate.getPvpDuration() / 100;
-			if (getEffector().getMaster() instanceof Player)
+			if (getEffector() instanceof Player) {
```
neon-dev: Please introduce a config property to make the old (non-retail) behavior optional.

game-server/src/com/aionemu/gameserver/configs/main/CustomConfig.java
```diff
@@ -263,4 +263,10 @@ public class CustomConfig {
 
 	@Property(key = "gameserver.rates.godstone.evaluation.cooldown_millis", defaultValue = "750")
 	public static int GODSTONE_EVALUATION_COOLDOWN_MILLIS;
+
+	/**
+	 * PvP config
```
neon-dev:
```suggestion
	 * Count summon-applied abnormal effects for cumulative resist.
```
Many *Config classes incorrectly use JavaDoc comments as section markers but since they're bound to individual declarations (as per spec) they should just describe the corresponding field's purpose.

game-server/src/com/aionemu/gameserver/controllers/effect/CumulativeResist.java
Birthright50: > I see now. The level shift does not affect the effective multipliers because this PR also changes `tryIncrementLevel()` to happen after calling `getDurationMultiplier()`. And the new default case should never be reached because on the fifth level, the spell will always be resisted.

In theory, the default case is reachable if the player has ARP (Abnormal Resist Penetration). In that case, it would behave like in the video example where I set AR to `0` while leaving `time_value `untouched. But normally, players don’t have ARP against fear, sleep, or paralysis.

neon-dev (changes requested):
~~I have noticed during my PTS tests that Curse of Fire and Curse of Water completely bypass the cumulative resist.
It needs more testing but my guess is that cumulative resists should only be checked for the first effect per skill, as these two fear skills the the only ones which have `Deform` instead of `Fear` as their first effect.
I haven't had the time yet to check if skills with secondary `Sleep` or `Paralyze` effects exist. If they do and they also bypass cumulative resistances, such a first-effect restriction could be implemented.~~

Edit: As it turns out, I was accidentally using a `noresist` version of the skill that is not available during regular gameplay.
Other available `noresist` skills, such as Curse of Old Roots, behave in the same way. Therefore, skills with the `noresist` flag must bypass the cumulative resistance.

game-server/src/com/aionemu/gameserver/skillengine/effect/EffectTemplate.java
```diff
@@ -516,7 +518,9 @@ public boolean calculateEffectResistRate(Effect effect, StatEnum statEnum) {
 		effectPower -= effected.getGameStats().getResistance(statEnum).getCurrent();
 
 		// calculate cumulative resist chance for fear, sleep and paralyze if effector and effected are players
-		if (effector.getMaster() instanceof Player && effected instanceof Player player)
+		boolean isEffectorPlayer =
+			CustomConfig.COUNT_SUMMON_EFFECTS_FOR_CUMULATIVE_RESIST ? effector.getMaster() instanceof Player : effector instanceof Player;
```
neon-dev:
```suggestion
		boolean isEffectorPlayer = (CustomConfig.COUNT_SUMMON_EFFECTS_FOR_CUMULATIVE_RESIST ? effector.getMaster() : effector) instanceof Player;
```

game-server/src/com/aionemu/gameserver/skillengine/model/Effect.java
```diff
@@ -845,14 +846,16 @@ private void stopPeriodicActions() {
 	private int calculateEffectsDuration() {
 		long duration = calculateTemplateDuration();
 
-		// adjust with pvp duration (not sure why some self target skills have pvp duration o.O idk how to handle that)
 		if (getEffected() instanceof Player effectedPlayer) {
-			if (skillTemplate.getPvpDuration() != 0 && !effector.equals(effected))
-				duration = duration * skillTemplate.getPvpDuration() / 100;
-			if (getEffector().getMaster() instanceof Player)
+			boolean isEffectorPlayer =
+				CustomConfig.COUNT_SUMMON_EFFECTS_FOR_CUMULATIVE_RESIST ? getEffector().getMaster() instanceof Player : getEffector() instanceof Player;
```
neon-dev:
```suggestion
			boolean isEffectorPlayer = (CustomConfig.COUNT_SUMMON_EFFECTS_FOR_CUMULATIVE_RESIST ? effector.getMaster() : effector) instanceof Player;
```

PR #138, opened 2026-05-14
Refactor stat caps and add dynamic limits

Birthright50:
## Summary

Refactor stat cap handling to support dynamic limits and adjust several caps to match retail behavior


neon-dev (commented):
I've cleaned this up for improved clarity.

Do `PHYSICAL_CRITICAL` and `MAGICAL_CRITICAL` have a lower limit of zero, by any chance?
If so, we should also replace all `registerDifferenceLimit()` invocations with `register()` and remove affected stats from line 36.

Birthright50:
> I've cleaned this up for improved clarity.
> 
> Do `PHYSICAL_CRITICAL` and `MAGICAL_CRITICAL` have a lower limit of zero, by any chance? If so, we should also replace all `registerDifferenceLimit()` invocations with `register()` and remove affected stats from line 36.

`PHYSICAL_CRITICAL `and `MAGICAL_CRITICAL `do have a lower limit of 0, but I didn’t quite understand why they should be removed from `registerDifferenceLimit`.

neon-dev:
> I didn’t quite understand why they should be removed from `registerDifferenceLimit`

My latest commit should make clear what I meant.

PR #143, opened 2026-06-04
feat: instance scaling

ghost:
adds optional instance scaling, scales NPC stats based on player count. configurable via instance.properties.

leaving the instance does not trigger a rescale,. this is a deliberate design choice intended to encourage teamwork and prevent abusing the mechanic.

### DIFF
--- a/game-server/config/main/instance.properties
+++ b/game-server/config/main/instance.properties
@@ -41,10 +41,6 @@ gameserver.instance.scaling.hp_floor = 0.5
 # Default: 0.5
 gameserver.instance.scaling.dmg_floor = 0.5
 
-# Minimum XP multiplier
-# Default: 0.5
-gameserver.instance.scaling.xp_floor = 0.5
-
 # Instance map IDs excluded from instance scaling
 gameserver.instance.scaling.excluded_maps =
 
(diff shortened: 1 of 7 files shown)


SVDNESS:
Issue: instance scaling is recalculated when a player enters the instance (`onEnterInstance`), but it is not recalculated when a player leaves.

Scenario:

1. A player enters the instance → NPC stats are scaled according to the current player count.
2. One or more players leave the instance.
3. NPC scaling is not updated because no recalculation is triggered.

As a result, NPC stats may no longer match the actual number of active players inside the instance. It is recommended to trigger `rescale()` when a player leaves the instance as well.

My opinion.

ghost:
its actually intentional design. not recalculating on leave was a deliberate choice to prevent players from exploiting the system by cycling members in/out to find the easiest configuration for a boss. rescaling on leave would effectively become a gameplay mechanic.

it also encourages players to stick together rather than manipulating group sizes...

SVDNESS:
Understood. If preventing scaling abuse is the intended design, then the current behavior is valid. My comment was based on the assumption that scaling should always reflect the current player count.

PR #145, opened 2026-06-12
Retail-accurate fear/confuse movement behavior.

SVDNESS:
Implements smooth NPC and player movement during fear and confuse effects,
matching retail behavior.

**Changes:**
- Flee destination updates every tick without stopping or resetting the current path
- Flee point is placed farther than the target can travel in one tick,
  so movement stays continuous and never stutters between ticks
- Eliminates the "run-stop-run" jitter caused by resetMove + moveToPoint on each tick


neon-dev:
I've changed `moveToPoint` so it can be used here. There were very little differences with the new `retargetPoint` method and it makes a lot of sense to allow moveToPoint to update the target coordinates, so now it can be used universally.

PR #147, opened 2026-06-15
Decrease the dispel counter only upon successful removal of a buff/debuff. 

Birthright50:
The counter should only decrease if the buff/debuff was actually dispelled. Otherwise, the power of the buff/debuff still decreases, but the search for the next 'candidate' for dispel continues.

### DIFF
--- a/game-server/src/com/aionemu/gameserver/controllers/effect/EffectController.java
+++ b/game-server/src/com/aionemu/gameserver/controllers/effect/EffectController.java
@@ -515,8 +515,8 @@ public class EffectController {
 							if (removePower(ef, power)) {
 								ef.setDesignatedDispelEffect(effect);
 								dispelledEffectCount++;
+								count--;
 							}
-							count--;
 						}
 						break;
 				}
@@ -579,10 +579,10 @@ public class EffectController {
 				if (remove) {
 					if (removePower(effect, power)) {
 						effectsToEnd.add(effect);
+						count--;
 					} else if (owner instanceof Player) {
 						insufficientDispelPower = true;
 					}
-					count--;
 				} else
 					insufficientDispelLevel = true;
 			}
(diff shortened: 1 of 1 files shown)


game-server/src/com/aionemu/gameserver/controllers/effect/EffectController.java
neon-dev, line 1: This would allow for an unlimited number of effects to have their power decreased, if the input power is not high enough to remove any effect. I don't think this is how it works on retail.
I think the correct solution would be to remove the counter, track the consumed power and break the loop once it reaches zero. In this case, the input power must **not** be `count * power`. Instead, everything must be wrapped in another loop, repeated by the number of `count`.

neon-dev (changes requested):

game-server/src/com/aionemu/gameserver/controllers/effect/EffectController.java
neon-dev: Thinking more about it, it is also possible that, when iterating the count, the search should continue starting from the last checked effect position (+1?).
Something like this:
```java
for (int i = 0; count > 0; count--) {
  for (int availablePower = power; availablePower > 0 && i < effects.size(); i++) {
    Effect effect = effects.get(i);
    int effectPower = effect.getPower();
    int remainingEffectPower = effect.removePower(availablePower); // TODO cap to zero in removePower
    int consumedEffectPower = effectPower - remainingEffectPower;
    availablePower -= consumedEffectPower;
    if (remainingEffectPower == 0) {
      effectsToEnd.add(effect);
    }
  }
}
```

Birthright50: > This would allow for an unlimited number of effects to have their power decreased, if the input power is not high enough to remove any effect. I don't think this is how it works on retail.
> I think the correct solution would be to remove the counter, track the consumed power and break the loop once it reaches zero. In this case, the input power must **not** be `count * power`. Instead, everything must be wrapped in another loop, repeated by the number of `count`.

Actually, that's exactly how it works on retail - if a buff/debuff cannot be dispelled, only its power is reduced, but the counter doesn't trigger. This continues until the counter reaches zero or there are no more candidates for dispel left.

PR #148, opened 2026-06-21
Fix incorrect skill mapping logic for req_dispel attributes

Birthright50:
In the previous [PR ](https://github.com/beyond-aion/aion-server/pull/139) introducing **req_dispel_count** and **req_dispel_level**, several issues were found in the data mapping logic between PTS versions. 

The mapping was performed only by skill `id ` using data from PTS 5.8, without validating the `name `field. This was based on the assumption that all skills from 4.8 exist in 5.8 with identical structure.
In practice, some skills were modified between versions (including changes in name), which led to incorrect parameters from 5.8 being applied to certain 4.8 skills.

A corrected matching strategy has been introduced:
If there is a full match of skill `id `+ `name `between 4.8 and 5.8 (with only grade differences allowed), parameters are taken from PTS 5.8.
If the `name `does not match while skill `id `does, a fallback is applied using PTS 4.6 with full skill `id `+ `name `validation.
[skill_flags.csv](https://github.com/user-attachments/files/29174641/skill_flags.csv)

### DIFF
--- a/game-server/data/static_data/skills/skill_templates.xml
+++ b/game-server/data/static_data/skills/skill_templates.xml
@@ -27460,7 +27460,7 @@
 		</effects>
 		<motion name="pointfire2" delay="10" />
 	</skill_template>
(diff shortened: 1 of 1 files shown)


neon-dev:
I have force-pushed a fixup to your commit that removes empty req_dispel_* attributes, which were causing a runtime error during startup.
Affected skills were 20762-20765, 21721 (test skills) and 21929 (npc skill).

PR #149, opened 2026-06-21
Implement support for missing skill attribute modifiers

Birthright50:
Missing skill attributes were added based on combined analysis of PTS 5.8 and 4.6 `skill_base.xml` and `skill_charge.xml`.

For intermediate calculation and validation purposes, a CSV dataset was created to consolidate and structure the comparison between PTS versions.
[skill_flags.csv](https://github.com/user-attachments/files/29180292/skill_flags.csv)

1) `apply_magical_critical` - same as the previous `applymcrit`, defaults to `false` when absent.

The flag decides **who may roll** the magical critical, not who may crit. The critical is rolled once per cast and target, and the following effect positions share the result.

| effect type | flag | behaviour |
|---|---|---|
| the six instant magical types (`SpellATK_Instant`, `SpellATKDrain_Instant`, `DelayedSpellATK_Instant`, `SignetBurst_Instant`, `ProcATK_Instant`) | read | **set**: reuses the critical of an earlier position, or rolls one and stores it for the later ones. **unset**: skipped entirely — no roll, no damage multiplier, and an already stored critical is ignored |
| periodic `SpellATK` / `SpellATKDrain` / `Poison` / `Bleed` | ignored | always reuses or rolls |
| `NoReduceSpellATK_Instant` | ignored | displays a critical that was already rolled, but never multiplies its own damage |
| `DispelBuffCounterATK` | ignored | never crits and never decides the critical of other positions |

With` E1 = SpellATK_Instant` and `E2 = SpellATK` (DoT), the flag controls whether the shared critical result is rolled or reused: when unset, `E1 `cannot crit and `E2 `rolls independently. when set, `E1 `rolls and `E2 `reuses its result. Verified on retail with `EL_ManaReverse_G6 (3555),` where the launched DoT can crit even when the flag is unset.

The shared result is stored before the resist check, while each effect applies its own critadddmg. Conditions/preeffects that remove an effect break the chain, and a DoT either crits on all ticks or none.

Added **critprobmod1**:` critprobmod2 + critprobmod1 * skillLvl.`
Fixed **critical_prob_mod2** values on 483 effects. Missing values were imported as 0, 376 effects should use the default 100, while 107 are explicit zeros.

The default is version-dependent and was verified on retail: 4.6 treats a missing critical_prob_mod2 as no critical, while 5.8 uses 100. Since 4.8 follows the newer skill system and our data aligns with 5.8, we use the 5.8 behavior.

2) `apply_casting_time_bonus `- removed the redundant hardcoded checks for effect type, avatar etc., since casting is governed solely by this flag. Also removed the check that cast speed bonuses only apply to magical skills - cast speed affects physical skills as well, for example https://aioncodex.com/48/skill/819/

For chargeable skills the bonus type comes from `charge_time_bonus_type` in `skill_charge.xml` instead of the caster's class (that check is gone):

   - `PHYSICAL` - attack speed always applies, this flag plays no role
   - `MAGICAL`- cast speed applies, but only when this flag is set
   - absent - the charge time is fixed.

      Charge skills only receive **half** of the speed bonus, so their caps are half of the normal ones: -37.5% for `MAGICAL` (half of the 75% casting time cap) and -25% / +50% for `PHYSICAL` (half of the 50% / 100% attack speed cap).

   `StatCapUtil `previously capped attack speed symmetrically at ±50% of base. Attack delay is now clamped to `[500, 10000]` and then to` [base * 0.5, base * 2]`, allowing slow effects to double the delay as on retail.
   The flag also controls whether cast speed affects both cast duration and motion time, so the hardcoded isMagical check was replaced with `skillTemplate.isApplyCastingTimeBonus()`.

3) `apply_heal_boost_bonus` - All possible heal skills were tested for how they are affected by the `apply_heal_boost_bonus `flag. The `DISEASE `abnormal effect applies to all of them.

   `CaseHeal`: deboost does not affect it, healboost does not affect it. The flag plays no role. No changes were made within this PR.

   `Heal_Instant` (percent=false, with flag true): deboost affects, healboost affects.
   `Heal_Instant` (percent=false, without boost flag): deboost affects, healboost does not affect.
   `Heal_Instant` (percent=true, without boost flag): deboost affects, healboost does not affect.
   `Heal_Instant` (percent=true, with flag true): deboost affects, healboost affects.

   `ProcHeal_instant` (percent=false, with flag true): deboost affects, healboost affects.
   `ProcHeal_instant` (percent=false, without boost flag): deboost affects, healboost does not affect.
   `ProcHeal_instant` (percent=true, without boost flag): deboost affects, healboost does not affect.
   `ProcHeal_instant` (percent=true, with flag true): deboost affects, healboost affects.

   `Heal` (percent=false, with flag true): deboost affects, healboost affects.
   `Heal` (percent=false, without boost flag): deboost does not affect, healboost does not affect.
   `Heal` (percent=true, without boost flag): deboost does not affect, healboost does not affect.
   `Heal` (percent=true, with flag true): deboost affects, healboost affects.

   For the `Heal` effect, the base heal amount is calculated at the moment the skill is applied (based on the caster's    healboost), while the actual healing amount changes dynamically depending on the deboost effect.
   Also, the `<= 0` check for periodic healing was removed, since the retail allows it.
   
4) `apply_magical_skill_boost_bonus` - logic using this flag has also been implemented, hardcoded instanceof checks have been removed from magic damage calculations (e.g., for `NoReduceSpellATKInstant`).


game-server/src/com/aionemu/gameserver/network/aion/serverpackets/SM_ATTACK_STATUS.java
```diff
@@ -145,7 +153,7 @@ protected void writeImpl(AionConnection con) {
 		writeC(type.getValue());
 		writeC(hpOrMp);
 		writeH(skillId);
-		writeH(logId);
+		writeH(criticalHit ? logId | CRITICAL_DISPLAY_CODE << 8 : logId); // the high byte of the log field is the display code
```
neon-dev:
```suggestion
		writeC(logId);
		writeC(criticalHit ? CRITICAL_DISPLAY_CODE : 0);
```
Please test to make sure I didn't mess up the byte order.

game-server/src/com/aionemu/gameserver/network/aion/serverpackets/SM_ATTACK_STATUS.java
```diff
@@ -171,7 +179,7 @@ protected void writeImpl(AionConnection con) {
 	 * heal_instant (regular) 171 protecteffect on protector - (8) 171 4.5
 	 * type="MP(21)" skillId="17722" logId="UNKNOWN(141) - mpattack
 	 * type="UNKNOWN(15)" skillId="2196" logId="UNKNOWN(112) - magiccounteratk
-	 * type="DAMAGE_HEAL_HP(7)" skillId="2858" logId="UNKNOWN(3073)" - spellatk(Flame Cage only)???
+	 * type="DAMAGE_HEAL_HP(7)" skillId="2858" logId="3073" - spellatk, 3073 = SPELLATK(1) | CRITICAL_DISPLAY_CODE << 8
```
neon-dev: I think this line can be removed.
```suggestion
```

game-server/src/com/aionemu/gameserver/skillengine/model/Skill.java
```diff
@@ -337,41 +337,36 @@ protected void updateCastDurationAndSpeed() {
...
-		int castDuration;
-		if (isPhysicalClass) // TODO check if attack speed should also affect magical classes
-			castDuration = (int) effector.getGameStats().getPositiveStat(StatEnum.ATTACK_SPEED, baseCastDuration);
-		else
-			castDuration = calculateMagicalCastDuration();
-		return Math.max(castDuration, (int) (baseCastDuration * 0.25f));
+		float speedRatio = switch (chargeTimeBonusType) {
+			case PHYSICAL -> {
+				Stat2 attackSpeed = effector.getGameStats().getAttackSpeed();
+				yield attackSpeed.getBase() == 0 ? 1f : (float) attackSpeed.getCurrent() / attackSpeed.getBase();
```
neon-dev: Does `getBase()` always return the correct, unmodified base for calculating the ratio?
The old `getPositiveStat()` approach seems more robust for this. Or something like 
```java
case PHYSICAL -> effector.getGameStats().getStat(StatEnum.ATTACK_SPEED, 1000f).get(Exact?)Current() / 1000f;
```

Birthright50: I think this could cause problems with effects where attack speed is set as a flat value rather than %.

neon-dev: Right, so I think ideally, the calculated base value in getAttackSpeed() should get its own method we can call here, so we can use it instead of 1000f.
https://github.com/beyond-aion/aion-server/blob/538b33d0774a042daae6de222db5e9208993439d/game-server/src/com/aionemu/gameserver/model/stats/container/PlayerGameStats.java#L87-L98

Birthright50: ~~I don't think I fully understand the idea, though. As far as I can tell, the current code in the branch already calculates the correct coefficient. There may have been some nuance with `baseRate`, but it's not actually set for attack speed by anyone. There could be an issue with `AbsoluteSlow` or Abs functions, but I haven't looked into that case, and I don't think it's really within the scope of this PR.~~
I think I get it now. I only skimmed it the first time. The idea is to get the base stat specifically as an `int` value so we don't have to use `1000`?

> Right, so I think ideally, the calculated base value in getAttackSpeed() should get its own method we can call here, so we can use it instead of 1000f.
> 
> https://github.com/beyond-aion/aion-server/blob/538b33d0774a042daae6de222db5e9208993439d/game-server/src/com/aionemu/gameserver/model/stats/container/PlayerGameStats.java#L87-L98

neon-dev: It's about using the same base value that's being used inside `getAttackSpeed()`. Using a fixed base of 1000 instead could calculate a wrong speedRatio, as you've already explained. The same is possible when using Stat2.getBase(), this value could have been modified by a StatFunction.

PR #153, opened 2026-07-03
Rework item use handling: casting delays, cancel bugs, quest dialogs, retail messages

Birthright50:
## Summary

This PR refactors item use flow to align it closer with retail behavior and removes a number of hardcoded casting delay values.

### Casting delays

- Added `casting_delay` to `item_templates.xml` using client 4.8 `client_items_etc.xml` and `client_items_misc.xml` data.
- Replaced hardcoded casting delays in multiple item actions with values from `item_templates.xml`.
- Added `casting_delay` support to `QuestStartAction` and `ReadAction` (8 items retail-confirmed to cast for 2s: 182207613, 182207865, 182213549, 182213550, 182215452, 182215453, 182215577, 182215578).
- Removed `casting_delay` support from `CompositionAction`, since retail ignores it even when present.
- The following actions technically support `casting_delay` on retail if defined in XML, but support was not added since no retail items currently use it: `CraftLearnAction`, `EmotionLearnAction`, `FireworksUseAction`, `TitleAddAction`, `SkillLearnAction`, `ExpandInventoryAction`.

### Cancel / interruption handling

- Added `ItemUseObserver` (or replaced `StartMovingListener`) for several actions to properly cancel ongoing item usage when another item is used. Previously this didn't always happen, which could delay execution (e.g. Soul Bind completing after its cast time) or leave items stuck unusable (e.g. enchant stones staying greyed out). Examples: `AnimationAddAction`, `EnchantItemAction`, `ItemSocketService`, `Equipment` (Soul Bind).
- Item actions that cast skills are now cancelled when another skill is used.
- Observer notification order changed to retail behavior: observers are notified before validating whether the item can actually be used (previously only before `act()`).
- Removed redundant `cancelUseItem()` calls from successful `act()` paths. Only observable difference found: `usingItem` stays set after successful completion until cleared elsewhere, which currently seems to affect only `CM_EMOTION`. Further verification welcome.

### Quest-start items

- Fixed quest dialog opening/closing immediately on use: the quest handler and `QuestStartAction` both tried to open it for the same use. Casting, message order, and race/level gating now match retail (checked against 4.6/5.8/4.8 client data).
- Dropped `queststart` from 4 items retail never actually wires to a quest (188100111, 188100112, 182212214, 182212215). Two are dead duplicates not in retail's item registry at all, and the other two are formally registered but have no activation and can't be used on retail either.

### System messages (retail wording/order)

- Corrected item-use system messages across most item actions to match retail wording/order: missing "You have used %0", wrong/reused message IDs, silent `canAct()` failures now notify the player, corrected cancel messages.
- Fixed message ordering for `ExtractAction`/`EnchantService.breakItem` (extract-success before acquired-item toast), `SkillLearnAction` (learned-skill toast before used-item, plus an explicit `STR_SKILL_LEARNED_NEW_SKILL` message the embedded `SM_SKILL_LIST` messageId doesn't actually trigger for these items), and `CraftLearnAction` (used-item before learned-recipe toast, confirmed the opposite order from `SkillLearnAction` on retail, not a universal rule).
- Fixed `DecomposeAction`'s regular (non-selectable) decompose path sending the wrong success message (`STR_DECOMPOSE_ITEM_SUCCEED`) instead of the retail-correct `STR_UNCOMPRESS_COMPRESSED_ITEM_SUCCEEDED`. `CM_SELECT_DECOMPOSABLE` (the selectable-decompose path) already used the right one.

### Multi-action item audit

Audited every item in the DB for co-occurring action tags. Only 3 combinations ever appear on the same item:

1. `queststart` + `read` (29 items): `ReadAction` was sending a redundant `STR_USE_ITEM`. It now no-ops if a sibling `QuestStartAction` exists on the item.
2. `charge` + `remodel` (1481 items): not a real conflict, `RemodelAction.canAct()` is hardcoded `return false`, so it's dead code that never runs.
3. `decompose` + `fireworkact` (4 event firecracker items): turned out to be a data bug, not a real combo. Cross-checked against the 5.8 server leak, all 4 items only have `<breakdown>1</breakdown>`, no firework field of any kind. Removed the erroneous `<fireworkact/>` tag from these 4 item templates (188051672, 188051711, 188052701, 188052702).

### ApExtractAction retail audit

tested every message and target-matching branch in `ApExtractAction` against 4.6 pts and fixed several bugs found in the process:

- `STR_MSG_AP_DECOMPOSE_WRONG_LEVEL` had the wrong parameters. Retail wording is "The level of %0 is below %1", where %0 is the extraction tool's name and %1 is the target item's name (both strings), but the code passed the target's name twice, once as a raw level number in place of %1.
- `UseTarget.ALL` was not handled at all and always failed with a wrong-category message. Confirmed on retail it acts as a universal target like `EQUIPMENT`.
- `UseTarget.WING` was never actually reachable due to dead code (a leftover check inside the `NONE` switch case that could never be true). AP extraction on wings was completely broken. Added a proper `WING` case.
- `UseTarget.OTHER` is used by retail for non-equipment ("junk") items with the `NONE` item group. Added handling for it, confirmed with a dedicated retail test item.
- Confirmed on retail that `EQUIPMENT` does not extend to `OTHER` category items, unlike `ALL`. Adjusted the code to match.
- Added the "bare" armor item groups (`TORSO`, `PANTS`, `SHOULDER`, `GLOVE`, `SHOES`, without a material prefix) to the armor category mapping. Confirmed on retail these are valid AP extraction targets that previously always failed with a wrong-category message.
- Added the missing casting delay (3 seconds, confirmed on retail, hardcoded like `PolishAction` since retail's own item data doesn't define it either) with proper cancel handling, reusing the `STR_MSG_AP_DECOMPOSE_ITEM_CANCELED` message that already existed in code but was never used.
- Added the missing check for extracting AP from an equipped item (`STR_MSG_AP_DECOMPOSE_WRONG_EQUIPED`, also previously defined but unused). Confirmed on retail this check takes priority over the other validation failures.

### Stale state during casting delays

While adding the `ApExtractAction` casting delay, found and fixed several other item actions that could apply a mutation based on state that changed during their own casting delay (target traded away, sold, re-equipped, or otherwise modified mid-cast), or that consumed items one by one and lost the earlier ones if a later one turned out to be missing:

- `AssemblyItemAction` now verifies all required parts are still present before consuming any of them, instead of consuming them one by one and losing the ones already consumed if a later part is missing.
- `EnchantService.breakItem` and `socketManastoneAct` now consume the cost item (the tool, or the manastone) before the item being acted on or before supplement materials, instead of the reverse order.
- `ChargeAction` now recomputes which equipped items to charge at the end of the cast instead of reusing a list captured at the start.
- `TuningAction` now re-validates the target item's presence and state before applying the tuning result.
- `InstanceTimeClear` now checks the cooldown is still there before consuming the item, instead of consuming it unconditionally.
- `ToyPetSpawnAction` now re-validates before spawning the kisk.

## Known issues (out of scope for this PR)

- ~30 quests have no handler implemented in this emulator at all (items behave correctly now, but there's nothing to accept/progress). Needs the quests themselves written from scratch.
- 188100111/112: shows a generic "can't use" message on use, where retail has zero reaction (verified on 4.6). This is a deliberate deviation, not an oversight: these items carry the same quest-activation scaffolding (`activate_target`, `use_delay`, quest-category flag) as real quest items from the same zone (Danuar Sanctuary), but aren't actually referenced as a work item by any quest in our data or in the retail quest.xml dump. They're orphaned, not wired to anything, likely just named or tagged from the same dev batch as quest 16982/26982's real quest items (182213547/182213548) without ever being hooked up. Silently swallowing the use attempt would make players wonder if the client lagged or the packet didn't go through. A clear "cannot be used" message is more helpful for genuinely dead items like these, and it costs nothing since there's no quest progress to preserve either way.
- Two of the 4 firecracker items above (188051672, 188051711) have no reward data in `decomposable_items.xml`, so `DecomposeAction.canAct()` fails outright ("cannot be decomposed") for them. Full decompose-item reward data reportedly wasn't present in the client before 4.9.

### DIFF
--- a/game-server/src/com/aionemu/gameserver/model/gameobjects/player/Equipment.java
+++ b/game-server/src/com/aionemu/gameserver/model/gameobjects/player/Equipment.java
@@ -740,7 +740,7 @@ public class Equipment implements Persistable {
 							new SM_ITEM_USAGE_ANIMATION(responder.getObjectId(), item.getObjectId(), item.getItemId(), 0, 8), true);
 					}
 				};
-				
+
 				responder.getObserveController().attach(observer);
 
 				// item usage animation
--- a/game-server/src/com/aionemu/gameserver/model/gameobjects/player/Equipment.java
+++ b/game-server/src/com/aionemu/gameserver/model/gameobjects/player/Equipment.java
@@ -8,8 +8,7 @@ import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.aionemu.gameserver.controllers.observer.ActionObserver;
-import com.aionemu.gameserver.controllers.observer.ObserverType;
+import com.aionemu.gameserver.controllers.observer.ItemUseObserver;
 import com.aionemu.gameserver.dao.InventoryDAO;
 import com.aionemu.gameserver.model.EmotionType;
 import com.aionemu.gameserver.model.Race;
@@ -731,26 +730,25 @@ public class Equipment implements Persistable {
 				PacketSendUtility.broadcastPacket(responder,
 					new SM_ITEM_USAGE_ANIMATION(responder.getObjectId(), item.getObjectId(), item.getItemId(), 5000, 4), true);
 
-				responder.getController().cancelTask(TaskId.ITEM_USE);
-
-				final ActionObserver moveObserver = new ActionObserver(ObserverType.MOVE) {
+				final ItemUseObserver observer = new ItemUseObserver() {
 
 					@Override
-					public void moved() {
-						responder.getController().cancelTask(TaskId.ITEM_USE);
+					public void abort() {
+						player.getController().cancelUseItem(false);
 						PacketSendUtility.sendPacket(responder, STR_SOUL_BOUND_ITEM_CANCELED(item.getL10n()));
 						PacketSendUtility.broadcastPacket(responder,
 							new SM_ITEM_USAGE_ANIMATION(responder.getObjectId(), item.getObjectId(), item.getItemId(), 0, 8), true);
 					}
 				};
-				responder.getObserveController().attach(moveObserver);
+				
+				responder.getObserveController().attach(observer);
 
 				// item usage animation
 				responder.getController().addTask(TaskId.ITEM_USE, ThreadPoolManager.getInstance().schedule(new Runnable() {
 
 					@Override
 					public void run() {
-						responder.getObserveController().removeObserver(moveObserver);
+						responder.getObserveController().removeObserver(observer);
 
 						PacketSendUtility.broadcastPacket(responder,
 							new SM_ITEM_USAGE_ANIMATION(responder.getObjectId(), item.getObjectId(), item.getItemId(), 0, 6), true);
--- a/game-server/src/com/aionemu/gameserver/model/gameobjects/player/Equipment.java
+++ b/game-server/src/com/aionemu/gameserver/model/gameobjects/player/Equipment.java
@@ -734,7 +734,7 @@ public class Equipment implements Persistable {
 
 					@Override
 					public void abort() {
-						player.getController().cancelUseItem(false);
+						responder.getController().cancelUseItem(false);
 						PacketSendUtility.sendPacket(responder, STR_SOUL_BOUND_ITEM_CANCELED(item.getL10n()));
 						PacketSendUtility.broadcastPacket(responder,
 							new SM_ITEM_USAGE_ANIMATION(responder.getObjectId(), item.getObjectId(), item.getItemId(), 0, 8), true);
--- a/game-server/src/com/aionemu/gameserver/model/gameobjects/player/Player.java
+++ b/game-server/src/com/aionemu/gameserver/model/gameobjects/player/Player.java
@@ -1009,6 +1009,14 @@ public class Player extends Creature {
 		return true;
 	}
 
+	public void startCooldown(Item item) {
+		ItemUseLimits limits = item.getItemTemplate().getUseLimits();
+		if (limits == null || limits.getDelayTime() <= 0 || hasCooldown(item))
+			return;
+
+		addItemCoolDown(limits.getDelayId(), System.currentTimeMillis() + limits.getDelayTime(), limits.getDelayTime() / 1000);
+	}
+
 	public long getItemReuseTime(int delayId) {
 		ItemCooldown cd = itemCoolDowns.get(delayId);
 		return cd == null ? 0 : cd.getReuseTime();
--- a/game-server/src/com/aionemu/gameserver/model/gameobjects/player/Player.java
+++ b/game-server/src/com/aionemu/gameserver/model/gameobjects/player/Player.java
@@ -1011,7 +1011,7 @@ public class Player extends Creature {
 
 	public void startCooldown(Item item) {
 		ItemUseLimits limits = item.getItemTemplate().getUseLimits();
-		if (limits == null || limits.getDelayTime() <= 0 || hasCooldown(item))
+		if (limits == null || limits.getDelayTime() <= 0)
 			return;
 
 		addItemCoolDown(limits.getDelayId(), System.currentTimeMillis() + limits.getDelayTime(), limits.getDelayTime() / 1000);
--- a/game-server/src/com/aionemu/gameserver/model/gameobjects/player/Player.java
+++ b/game-server/src/com/aionemu/gameserver/model/gameobjects/player/Player.java
@@ -1575,7 +1575,7 @@ public class Player extends Creature {
 		this.isInSprintMode = isInSprintMode;
 	}
 
-	public void setRideObservers(ActionObserver observer) {
+	public void addRideObserver(ActionObserver observer) {
 		if (rideObservers == null)
 			rideObservers = new ArrayList<>();
 
--- a/game-server/src/com/aionemu/gameserver/model/gameobjects/player/PortalCooldownList.java
+++ b/game-server/src/com/aionemu/gameserver/model/gameobjects/player/PortalCooldownList.java
@@ -59,8 +59,10 @@ public class PortalCooldownList {
 	 */
 	public PortalCooldown getOrCreatePortalCooldown(int worldId) {
 		long reuseTime = DataManager.INSTANCE_COOLTIME_DATA.calculateInstanceEntranceCooltime(owner, worldId);
-		if (reuseTime == 0)
-			return null;
+		return reuseTime == 0 ? null : getOrCreatePortalCooldown(worldId, reuseTime);
+	}
+
+	private synchronized PortalCooldown getOrCreatePortalCooldown(int worldId, long reuseTime) {
 		if (portalCooldowns == null)
 			portalCooldowns = new HashMap<>();
 		return portalCooldowns.computeIfAbsent(worldId, id -> new PortalCooldown(id, reuseTime, 0));
@@ -75,15 +77,7 @@ public class PortalCooldownList {
 	}
 
 	public void addPortalCooldown(int worldId, long useDelay) {
-		if (portalCooldowns == null)
-			portalCooldowns = new HashMap<>();
-
-		PortalCooldown portalCooldown = portalCooldowns.get(worldId);
-		if (portalCooldown == null)
-			portalCooldown = new PortalCooldown(worldId, useDelay, 0);
-
-		portalCooldown.increaseEnterCount();
-		portalCooldowns.put(worldId, portalCooldown);
+		getOrCreatePortalCooldown(worldId, useDelay).increaseEnterCount();
 
 		PortalCooldownsDAO.storePortalCooldowns(owner);
 
--- a/game-server/src/com/aionemu/gameserver/model/gameobjects/player/PortalCooldownList.java
+++ b/game-server/src/com/aionemu/gameserver/model/gameobjects/player/PortalCooldownList.java
@@ -53,6 +53,19 @@ public class PortalCooldownList {
 		return portalCooldowns == null ? null : portalCooldowns.get(worldId);
 	}
 
+	/**
+	 * @return The cooldown of the given world, creating it if absent, or null if the world has no entry limit (the client
+	 *         displays such instances as unlimited, as long as we don't send any entry info for them).
+	 */
+	public PortalCooldown getOrCreatePortalCooldown(int worldId) {
+		long reuseTime = DataManager.INSTANCE_COOLTIME_DATA.calculateInstanceEntranceCooltime(owner, worldId);
+		if (reuseTime == 0)
+			return null;
+		if (portalCooldowns == null)
+			portalCooldowns = new HashMap<>();
+		return portalCooldowns.computeIfAbsent(worldId, id -> new PortalCooldown(id, reuseTime, 0));
+	}
+
 	public Map<Integer, PortalCooldown> getPortalCoolDowns() {
 		return portalCooldowns;
 	}
--- a/game-server/src/com/aionemu/gameserver/model/templates/item/actions/AnimationAddAction.java
+++ b/game-server/src/com/aionemu/gameserver/model/templates/item/actions/AnimationAddAction.java
@@ -65,6 +65,7 @@ public class AnimationAddAction extends AbstractItemAction {
 			new SM_ITEM_USAGE_ANIMATION(player.getObjectId(), parentItem.getObjectId(), parentItem.getItemTemplate().getTemplateId(), castingDelay, 0, 0));
 		player.getController().addTask(TaskId.ITEM_USE, ThreadPoolManager.getInstance().schedule(() -> {
 			player.getObserveController().removeObserver(observer);
+			player.startCooldown(parentItem);
 			finishUse(player, parentItem);
 		}, castingDelay));
 	}
--- a/game-server/src/com/aionemu/gameserver/model/templates/item/actions/AnimationAddAction.java
+++ b/game-server/src/com/aionemu/gameserver/model/templates/item/actions/AnimationAddAction.java
@@ -5,6 +5,7 @@ import javax.xml.bind.annotation.XmlAccessorType;
 import javax.xml.bind.annotation.XmlAttribute;
 import javax.xml.bind.annotation.XmlType;
 
+import com.aionemu.gameserver.controllers.observer.ItemUseObserver;
 import com.aionemu.gameserver.model.TaskId;
 import com.aionemu.gameserver.model.gameobjects.Item;
 import com.aionemu.gameserver.model.gameobjects.player.Player;
@@ -38,41 +39,57 @@ public class AnimationAddAction extends AbstractItemAction {
 			PacketSendUtility.sendPacket(player, SM_SYSTEM_MESSAGE.STR_ITEM_COLOR_ERROR());
 			return false;
 		}
-
 		return true;
 	}
 
 	@Override
 	public void act(final Player player, final Item parentItem, Item targetItem, Object... params) {
-		player.getController().cancelUseItem();
-		PacketSendUtility.sendPacket(player, new SM_ITEM_USAGE_ANIMATION(player.getObjectId(), parentItem.getObjectId(), parentItem.getItemTemplate()
-			.getTemplateId(), 1000, 0, 0));
-		player.getController().addTask(TaskId.ITEM_USE, ThreadPoolManager.getInstance().schedule(new Runnable() {
+		int castingDelay = parentItem.getItemTemplate().getCastingDelay();
+		if (castingDelay <= 0) {
+			finishUse(player, parentItem);
+			return;
+		}
+		final ItemUseObserver observer = new ItemUseObserver() {
 
 			@Override
-			public void run() {
-				if (player.getInventory().decreaseItemCount(parentItem, 1) != 0)
-					return;
-				if (idle != null) {
-					addMotion(player, idle);
-				}
-				if (run != null) {
-					addMotion(player, run);
-				}
-				if (jump != null) {
-					addMotion(player, jump);
-				}
-				if (rest != null) {
-					addMotion(player, rest);
-				}
-				if (shop != null) {
-					addMotion(player, shop);
-				}
-				PacketSendUtility.broadcastPacketAndReceive(player,
-					new SM_ITEM_USAGE_ANIMATION(player.getObjectId(), parentItem.getObjectId(), parentItem.getItemId(), 0, 1, 0));
-				PacketSendUtility.broadcastPacket(player, new SM_MOTION(player.getObjectId(), player.getMotions().getActiveMotions()), false);
+			public void abort() {
+				player.getController().cancelUseItem();
+				PacketSendUtility.sendPacket(player, SM_SYSTEM_MESSAGE.STR_ITEM_CANCELED());
+				player.getObserveController().removeObserver(this);
 			}
-		}, 1000));
+
+		};
+
+		player.getObserveController().attach(observer);
+		PacketSendUtility.sendPacket(player,
+			new SM_ITEM_USAGE_ANIMATION(player.getObjectId(), parentItem.getObjectId(), parentItem.getItemTemplate().getTemplateId(), castingDelay, 0, 0));
+		player.getController().addTask(TaskId.ITEM_USE, ThreadPoolManager.getInstance().schedule(() -> {
+			player.getObserveController().removeObserver(observer);
+			finishUse(player, parentItem);
+		}, castingDelay));
+	}
+
+	private void finishUse(Player player, Item parentItem) {
+		if (player.getInventory().decreaseItemCount(parentItem, 1) != 0)
+			return;
+		if (idle != null) {
+			addMotion(player, idle);
+		}
+		if (run != null) {
+			addMotion(player, run);
+		}
+		if (jump != null) {
+			addMotion(player, jump);
+		}
+		if (rest != null) {
+			addMotion(player, rest);
+		}
+		if (shop != null) {
+			addMotion(player, shop);
+		}
+		PacketSendUtility.broadcastPacketAndReceive(player,
+			new SM_ITEM_USAGE_ANIMATION(player.getObjectId(), parentItem.getObjectId(), parentItem.getItemId(), 0, 1, 0));
+		PacketSendUtility.broadcastPacket(player, new SM_MOTION(player.getObjectId(), player.getMotions().getActiveMotions()), false);
 	}
 
 	private void addMotion(Player player, int motionId) {
--- a/game-server/src/com/aionemu/gameserver/model/templates/item/actions/AnimationAddAction.java
+++ b/game-server/src/com/aionemu/gameserver/model/templates/item/actions/AnimationAddAction.java
@@ -65,7 +65,6 @@ public class AnimationAddAction extends AbstractItemAction {
 			new SM_ITEM_USAGE_ANIMATION(player.getObjectId(), parentItem.getObjectId(), parentItem.getItemTemplate().getTemplateId(), castingDelay, 0, 0));
 		player.getController().addTask(TaskId.ITEM_USE, ThreadPoolManager.getInstance().schedule(() -> {
 			player.getObserveController().removeObserver(observer);
-			player.startCooldown(parentItem);
 			finishUse(player, parentItem);
 		}, castingDelay));
 	}
@@ -73,6 +72,7 @@ public class AnimationAddAction extends AbstractItemAction {
 	private void finishUse(Player player, Item parentItem) {
 		if (player.getInventory().decreaseItemCount(parentItem, 1) != 0)
 			return;
+		player.startCooldown(parentItem);
 		if (idle != null) {
 			addMotion(player, idle);
 		}
--- a/game-server/src/com/aionemu/gameserver/model/templates/item/actions/AnimationAddAction.java
+++ b/game-server/src/com/aionemu/gameserver/model/templates/item/actions/AnimationAddAction.java
@@ -87,6 +87,7 @@ public class AnimationAddAction extends AbstractItemAction {
 		if (shop != null) {
 			addMotion(player, shop);
 		}
+		PacketSendUtility.sendPacket(player, SM_SYSTEM_MESSAGE.STR_USE_ITEM(parentItem.getL10n()));
 		PacketSendUtility.broadcastPacketAndReceive(player,
 			new SM_ITEM_USAGE_ANIMATION(player.getObjectId(), parentItem.getObjectId(), parentItem.getItemId(), 0, 1, 0));
 		PacketSendUtility.broadcastPacket(player, new SM_MOTION(player.getObjectId(), player.getMotions().getActiveMotions()), false);
--- a/game-server/src/com/aionemu/gameserver/model/templates/item/actions/ApExtractAction.java
+++ b/game-server/src/com/aionemu/gameserver/model/templates/item/actions/ApExtractAction.java
@@ -144,13 +144,14 @@ public class ApExtractAction extends AbstractItemAction {
 		player.getObserveController().attach(observer);
 		player.getController().addTask(TaskId.ITEM_USE, ThreadPoolManager.getInstance().schedule(() -> {
 			player.getObserveController().removeObserver(observer);
-			player.startCooldown(parentItem);
 			finishUse(player, parentItem, targetItem);
 		}, CASTING_DELAY));
 	}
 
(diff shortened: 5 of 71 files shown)


game-server/src/com/aionemu/gameserver/model/templates/item/actions/QuestStartAction.java
```diff
@@ -30,21 +34,60 @@ public class QuestStartAction extends AbstractItemAction {
...
+
+		// Retail only gates "doc_"-named quest items on race/level/etc.; plain quest items skip this check
+		// entirely and always reach the accept dialog (confirmed on retail 5.8)
+		boolean isDocument = item.getItemTemplate().getClientName().toLowerCase().startsWith("doc_");
+		boolean canStart = !alreadyActive && (!isDocument || QuestService.checkStartConditions(player, questid, true, 0, true, false, false));
+
+		PacketSendUtility.sendPacket(player, SM_SYSTEM_MESSAGE.STR_USE_ITEM(item.getL10n()));
+
+		if (!canStart)
+			return; // quest already active, or race/level/etc. requirements not met (checkStartConditions already sent the message)
```
neon-dev: This will not send a failure message for documents if the quest is already active and I don't think it's right to skip the quest repeat count check for documents.

game-server/src/com/aionemu/gameserver/skillengine/model/Skill.java
```diff
@@ -353,8 +353,10 @@ private int calculateChargeCastDuration() {
 	}
 
 	private int calculateCastDuration() {
-		// ap & cash revival stones, or 2nd+ time of multicast-skill activation
-		if (getSkillId() == 10802 || getMultiCastCount() > 0)
+		if (this.getItemTemplate() != null)
+			return this.getItemTemplate().getCastingDelay();
```
neon-dev:
```suggestion
		if (getItemTemplate() != null)
			return getItemTemplate().getCastingDelay();
```

game-server/src/com/aionemu/gameserver/services/EnchantService.java
```diff
@@ -66,9 +66,11 @@ else if (rndEffectiveLevel >= calculateEffectiveLevel(EnchantmentStone.BETA))
 		else
 			stoneId = 166000191; // Alpha
 
-		if (inventory.delete(targetItem) != null) {
-			if (inventory.decreaseByObjectId(parentItem.getObjectId(), 1))
+		if (inventory.decreaseByObjectId(parentItem.getObjectId(), 1)) {
+			if (inventory.delete(targetItem) != null) {
+				PacketSendUtility.sendPacket(player, SM_SYSTEM_MESSAGE.STR_DECOMPOSE_ITEM_SUCCEED(targetItem.getL10n()));
 				ItemService.addItem(player, stoneId, itemTemplate.isWeapon() ? Rnd.get(2, 5) : Rnd.get(1, 3));
+			}
```
neon-dev:
```suggestion
		if (inventory.decreaseByObjectId(parentItem.getObjectId(), 1) && inventory.delete(targetItem) != null) {
			ItemService.addItem(player, stoneId, itemTemplate.isWeapon() ? Rnd.get(2, 5) : Rnd.get(1, 3));
			PacketSendUtility.sendPacket(player, SM_SYSTEM_MESSAGE.STR_DECOMPOSE_ITEM_SUCCEED(targetItem.getL10n()));
```
Retail adds the item before sending the success message and the AuditLogger should be called from both failure conditions.

game-server/src/com/aionemu/gameserver/network/aion/serverpackets/SM_ITEM_USAGE_ANIMATION.java
neon-dev, line 1: I assume this file was committed unintentionally.

game-server/src/com/aionemu/gameserver/model/gameobjects/player/PortalCooldownList.java
```diff
@@ -53,6 +53,19 @@ public PortalCooldown getPortalCooldown(int worldId) {
...
+	 * @return The cooldown of the given world, creating it if absent, or null if the world has no entry limit (the client
+	 *         displays such instances as unlimited, as long as we don't send any entry info for them).
+	 */
+	public PortalCooldown getOrCreatePortalCooldown(int worldId) {
+		long reuseTime = DataManager.INSTANCE_COOLTIME_DATA.calculateInstanceEntranceCooltime(owner, worldId);
+		if (reuseTime == 0)
+			return null;
+		if (portalCooldowns == null)
+			portalCooldowns = new HashMap<>();
+		return portalCooldowns.computeIfAbsent(worldId, id -> new PortalCooldown(id, reuseTime, 0));
```
neon-dev, line 68: Please create a helper method from this and reuse it in `addPortalCooldown`. You could also synchronize over `this` to prevent race conditions. I know this won't solve all potential race conditions with this class but I also don't think it's crucial to rewrite the whole class either.

game-server/src/com/aionemu/gameserver/model/templates/item/actions/QuestStartAction.java
```diff
@@ -30,21 +34,60 @@ public class QuestStartAction extends AbstractItemAction {
...
+
+	private void finishUse(Player player, Item item) {
+		PacketSendUtility.broadcastPacketAndReceive(player, new SM_ITEM_USAGE_ANIMATION(player.getObjectId(), item.getObjectId(), item.getItemId()));
+
+		QuestState qs = player.getQuestStateList().getQuestState(questid);
+		boolean alreadyActive = qs != null && (qs.getStatus() == QuestStatus.START || qs.getStatus() == QuestStatus.REWARD);
+
+		// Retail only gates "doc_"-named quest items on race/level/etc.; plain quest items skip this check
+		// entirely and always reach the accept dialog (confirmed on retail 5.8)
+		boolean isDocument = item.getItemTemplate().getClientName().toLowerCase().startsWith("doc_");
```
neon-dev: Is it really the "doc_" condition on retail that controls this and not the presence of a `ReadAction`?
I've found a few "doc_" items which don't have a `ReadAction` you can test this with: `doc_quest_11212a`, `doc_quest_11216a`, `doc_quest_2918a`.

If the "doc_" check could be removed, we could also save a good chunk of memory by removing the `clientName` field from `ItemTemplate.java` again.
If, on the other hand, the field needs to stay, please remove the file "data/handlers/consolecommands/data/items.xml" and replace all its usages.

game-server/src/com/aionemu/gameserver/restrictions/PlayerRestrictions.java
```diff
@@ -65,8 +65,14 @@ public static boolean canUseSkill(Player player, Skill skill) {
...
-		if (player.isCasting())
-			return false;
+		if (player.isCasting()) {
+			if (player.getCastingSkill().getItemTemplate() == null) {
+				return false;
+			} else {
+				player.getController().cancelCurrentSkill(null);
+			}
+		}
+
```
neon-dev:
```suggestion
		if (player.isCasting() && player.getCastingSkill().getItemTemplate() == null)
			return false;
```
`cancelCurrentSkill()` should not be called from PlayerRestrictions but in the PlayerController and I think the comments are obsolete.

game-server/src/com/aionemu/gameserver/model/gameobjects/player/Equipment.java
```diff
@@ -731,26 +730,25 @@ public void acceptRequest(Player requester, Player responder) {
...
-				responder.getController().cancelTask(TaskId.ITEM_USE);
-
-				final ActionObserver moveObserver = new ActionObserver(ObserverType.MOVE) {
+				final ItemUseObserver observer = new ItemUseObserver() {
 
 					@Override
-					public void moved() {
-						responder.getController().cancelTask(TaskId.ITEM_USE);
+					public void abort() {
+						player.getController().cancelUseItem(false);
```
neon-dev:
```suggestion
						responder.getController().cancelUseItem(false);
```

game-server/src/com/aionemu/gameserver/model/templates/item/actions/ApExtractAction.java
```diff
@@ -83,31 +105,74 @@ public boolean canAct(Player player, Item parentItem, Item targetItem, Object...
...
+					SM_SYSTEM_MESSAGE.STR_MSG_AP_DECOMPOSE_WRONG_TARGET_ITEM_CATEGORY(parentItem.getL10n(), targetItem.getL10n()));
 				return false;
 		}
-		return (target == UseTarget.EQUIPMENT || target == type);
+		// EQUIPMENT is a shorthand for "any of WEAPON/ARMOR/ACCESSORY/WING", confirmed on retail it does NOT also match OTHER
+		if (target == UseTarget.ALL || target == type || (target == UseTarget.EQUIPMENT && type != UseTarget.OTHER))
+			return true;
+		PacketSendUtility.sendPacket(player,
+			SM_SYSTEM_MESSAGE.STR_MSG_AP_DECOMPOSE_WRONG_TARGET_ITEM_CATEGORY(parentItem.getL10n(), targetItem.getL10n()));
+		return false;
```
neon-dev:
```suggestion
		if (target != UseTarget.ALL && target != type && !(target == UseTarget.EQUIPMENT && type != UseTarget.OTHER)) {
			PacketSendUtility.sendPacket(player,
			SM_SYSTEM_MESSAGE.STR_MSG_AP_DECOMPOSE_WRONG_TARGET_ITEM_CATEGORY(parentItem.getL10n(), targetItem.getL10n()));
			return false;
		}
		return true;
```
Let's flip this around to keep things consistent.

game-server/src/com/aionemu/gameserver/model/templates/item/actions/ApExtractAction.java
```diff
@@ -83,31 +105,74 @@ public boolean canAct(Player player, Item parentItem, Item targetItem, Object...
...
-			if (inventory.decreaseByObjectId(parentItem.getObjectId(), 1))
-				AbyssPointsService.addAp(player, ap);
-		} else
+		if (inventory.getItemByObjId(parentItem.getObjectId()) == null || inventory.getItemByObjId(targetItem.getObjectId()) == null) {
 			AuditLogger.log(player, "possibly using item AP extraction hack");
+			return false;
+		}
+		int ap = (int) (acquisition.getRequiredAp() * rate);
+		inventory.decreaseByObjectId(parentItem.getObjectId(), 1);
+		inventory.delete(targetItem);
```
neon-dev:
```suggestion
		if (!inventory.decreaseByObjectId(parentItem.getObjectId(), 1) || inventory.delete(targetItem) == null) {
			AuditLogger.log(player, "possibly using item AP extraction hack");
			return false;
		}
		int ap = (int) (acquisition.getRequiredAp() * rate);
```

neon-dev (changes requested):

game-server/src/com/aionemu/gameserver/services/EnchantService.java
Birthright50: I can’t confirm this. On 4.6, it doesn’t show how many items were added at all. On 5.8, I got the following: 
image

game-server/src/com/aionemu/gameserver/model/gameobjects/player/Equipment.java
Birthright50: basically, it’s not really a bug since they’re always the same (player and responder), but I fixed it as well

game-server/src/com/aionemu/gameserver/model/templates/item/actions/QuestStartAction.java
Birthright50: Retail sends no failure message in either case, and doesn't open the accept dialog either. With the quest already active, using 182204501 gives just `You have used Lepharist Book.` plus the read window. Same after completing quest 2670 (`max_repeat_count="1"`) and using the item again. [Lepharist Book](https://www.youtube.com/watch?v=TOeUJ6H5ej0)

game-server/src/com/aionemu/gameserver/model/templates/item/actions/QuestStartAction.java
Birthright50: Tested on an asmo 65 lvl:
182215139 | queststart | levels 40-49 | `You can only receive this quest when your level is 49 or below.` -> `You have used …` 

182213209 | queststart | levels 21-23  | `You can only receive this quest when your level is 23 or below.` -> `You have used …` 

182201304 | queststart | elyos only | `Your race cannot receive this quest.` -> `You have used …`

182201770 | queststart + read |elyos only  | same, and the read window still opens

The first row is a document without a `ReadAction`, the second and third are plain quest items, so neither the "doc_" name nor the presence of a `ReadAction` decides it. `checkStartConditions` is called unconditionally now and `ItemTemplate.clientName` is gone again, so "data/handlers/consolecommands/data/items.xml" stays as it is.
The plain quest items I originally tested happened to lead to quests the test character was eligible for, so there was nothing for the server to complain about. I read that silence as "plain items are never checked", when in fact the difference was in the quest requirements, not in the item.

game-server/src/com/aionemu/gameserver/services/EnchantService.java
Birthright50: > Retail adds the item before sending the success message and the AuditLogger should be called from both failure conditions.

Figured out the ordering: when the reward merges into a stack you already have, retail sends the inventory packet before the "opened/used" messages, when it creates a new inventory slot it sends it after. Used one box 9 times in a row -every first occurrence of an item type printed the acquisition last, every repeat printed it first. Would it be worth changing the order just for this? The rule is pretty strange.

neon-dev: Got it, in this case there's no reason to change the order. Since you already corrected the logging condition, I'll mark this conversation as resolved.

game-server/src/com/aionemu/gameserver/model/gameobjects/player/Equipment.java
neon-dev: Correct. This suggestion was just to keep things consistent with the other lines in this method. 😄

game-server/src/com/aionemu/gameserver/model/templates/item/actions/QuestStartAction.java
```diff
@@ -30,21 +33,57 @@ public class QuestStartAction extends AbstractItemAction {
...
+			return;
+		}
+		PacketSendUtility.broadcastPacket(player,
+			new SM_ITEM_USAGE_ANIMATION(player.getObjectId(), parentItem.getObjectId(), parentItem.getItemId(), castingDelay, 0, 1), true);
+		final ItemUseObserver observer = new ItemUseObserver() {
+
+			@Override
+			public void abort() {
+				player.getController().cancelUseItem(false);
+				player.removeItemCoolDown(parentItem.getItemTemplate().getUseLimits().getDelayId());
```
neon-dev: Why is there a cooldown to cancel? Shouldn't the cooldown only start after the cast animation has finished?
The same line exists in several other *Action classes as well.

game-server/src/com/aionemu/gameserver/model/templates/item/actions/ApExtractAction.java
```diff
@@ -2,36 +2,53 @@
...
+		}
+		if (parentItem.getItemTemplate().getLevel() < targetItem.getItemTemplate().getLevel()) {
+			PacketSendUtility.sendPacket(player, SM_SYSTEM_MESSAGE.STR_MSG_AP_DECOMPOSE_WRONG_LEVEL(parentItem.getL10n(), targetItem.getL10n()));
 			return false;
+		}
+		if (parentItem.getItemTemplate().getItemQuality() != targetItem.getItemTemplate().getItemQuality()) {
+			PacketSendUtility.sendPacket(player,
+				SM_SYSTEM_MESSAGE.STR_MSG_AP_DECOMPOSE_WRONG_QUALITY(parentItem.getL10n(), targetItem.getL10n()));
+			return false;
+		}
```
neon-dev: The message text indicates that retail could be checking `<` here. Do you know if this is the case?

game-server/src/com/aionemu/gameserver/model/templates/item/actions/DecomposeAction.java
```diff
@@ -129,20 +126,23 @@ public void act(final Player player, final Item parentItem, final Item targetIte
...
+			new SM_ITEM_USAGE_ANIMATION(player.getObjectId(), parentItem.getObjectId(), parentItem.getItemId(), castingDelay, 0, 0), true);
 
 		final ItemUseObserver observer = new ItemUseObserver() {
 
 			@Override
 			public void abort() {
-				player.getController().cancelTask(TaskId.ITEM_USE);
+				player.getController().cancelUseItem(false);
 				player.removeItemCoolDown(parentItem.getItemTemplate().getUseLimits().getDelayId());
 				PacketSendUtility.sendPacket(player, SM_SYSTEM_MESSAGE.STR_DECOMPOSE_ITEM_CANCELED(parentItem.getL10n()));
```
neon-dev: Since `finishUse` now sends the `STR_UNCOMPRESS_COMPRESSED_ITEM_SUCCEEDED` success message, shouldn't we send `STR_UNCOMPRESS_COMPRESSED_ITEM_CANCELED` here?

game-server/src/com/aionemu/gameserver/model/templates/item/actions/SkillLearnAction.java
```diff
@@ -51,16 +53,18 @@ public boolean canAct(Player player, Item parentItem, Item targetItem, Object...
...
+			new SM_ITEM_USAGE_ANIMATION(player.getObjectId(), parentItem.getObjectId(), parentItem.getItemId()), true);
 
-		// add skill
+		// add skill; SM_SKILL_LIST's embedded messageId alone doesn't show the "you learned" toast for these items, so send it explicitly
 		SkillLearnService.learnSkillBook(player, skillid);
 
+		SkillTemplate skillTemplate = DataManager.SKILL_DATA.getSkillTemplate(skillid);
+		PacketSendUtility.sendPacket(player,
+			SM_SYSTEM_MESSAGE.STR_SKILL_LEARNED_NEW_SKILL(skillTemplate.getL10n(), String.valueOf(skillTemplate.getLvl())));
+
```
neon-dev: For me, the client displays this message from `SM_SKILL_LIST` already, so this change makes it appear twice in chat.
Tested with item ID 169500932.

game-server/src/com/aionemu/gameserver/model/templates/item/actions/SkillUseAction.java
```diff
@@ -100,7 +100,6 @@ private static boolean isIneffectiveHealSkill(List<EffectTemplate> effects, List
 	public void act(Player player, Item parentItem, Item targetItem, Object... params) {
 		Skill skill = SkillEngine.getInstance().getSkill(player, skillid, level, player.getTarget(), parentItem.getItemTemplate());
 		if (skill != null) {
-			player.getController().cancelUseItem();
```
neon-dev, line 103: Can't we just leave this here as a precaution? While `CM_CASTSPELL` cancels item usages already, I'm not sure if this action could be also triggered from other places.
Or maybe `useSkill()` should handle it?

neon-dev (changes requested):

game-server/src/com/aionemu/gameserver/model/templates/item/actions/SkillUseAction.java
Birthright50: `SkillUseAction.act()` has one caller - `CM_USE_ITEM`. 
And `CM_USE_ITEM `already cancels a running item cast twice before reaching act():
1) `if (player.isCasting()) cancelCurrentSkill(null)`
2)  `notifyItemuseObservers(item)` -> every `ItemUseObserver.abort() -> cancelUseItem(false)`

`PlayerController.useSkill()` already does exactly that (see the` player.getCastingSkill().getItemTemplate() != null` block).

game-server/src/com/aionemu/gameserver/model/templates/item/actions/ApExtractAction.java
Birthright50: Yes, retail checks <. I hadn’t tested this case before.
And two more things (after analyzing the binary): 
1) a category mismatch sends `STR_MSG_AP_DECOMPOSE_CANNOT`, not `STR_MSG_AP_DECOMPOSE_WRONG_TARGET_ITEM_CATEGORY`.
2)  `STR_MSG_AP_DECOMPOSE_WRONG_EQUIPED `is absent server-side, so that one is presumably client-side. I left our check in, it's harmless.

game-server/src/com/aionemu/gameserver/model/templates/item/actions/DecomposeAction.java
Birthright50: Right, changed.

game-server/src/com/aionemu/gameserver/model/templates/item/actions/SkillLearnAction.java
Birthright50: Removed, but the actual bug is elsewhere though. When the learned skill has a skillLearn predecessor that is already known,` PlayerSkillList.addSkill()` marks it `isNew = false`, so `SkillLearnService.sendPacket()` sends `SM_SKILL_LIST(skill, 0)` and the client stays silent. On retail 4.6 the message is shown in that case - e.g. Transformation: White Tiger II (169500917) with tier I already learned. also `autoLearnSkills()`adds every tier as its own skill id, so learning tier III with tier II known gets silenced the same way.

game-server/src/com/aionemu/gameserver/model/templates/item/actions/AnimationAddAction.java
```diff
@@ -65,6 +65,7 @@ public void abort() {
 			new SM_ITEM_USAGE_ANIMATION(player.getObjectId(), parentItem.getObjectId(), parentItem.getItemTemplate().getTemplateId(), castingDelay, 0, 0));
 		player.getController().addTask(TaskId.ITEM_USE, ThreadPoolManager.getInstance().schedule(() -> {
 			player.getObserveController().removeObserver(observer);
+			player.startCooldown(parentItem);
 			finishUse(player, parentItem);
```
neon-dev: `startCooldown` should be called from `finishUse` where such a method exists, so items without a `castingDelay` are also covered.

game-server/src/com/aionemu/gameserver/model/gameobjects/player/Player.java
```diff
@@ -1009,6 +1009,14 @@ public boolean hasCooldown(Item item) {
 		return true;
 	}
 
+	public void startCooldown(Item item) {
+		ItemUseLimits limits = item.getItemTemplate().getUseLimits();
+		if (limits == null || limits.getDelayTime() <= 0 || hasCooldown(item))
```
neon-dev: Why return on hasCooldown, shouldn't we just update the cooldown in any case?

game-server/src/com/aionemu/gameserver/model/templates/item/actions/ApExtractAction.java
neon-dev: 1. Now the old message is unused. Maybe we're missing some check?
2. Not only is it harmless but it's generally also important to check everything on the server side as well, to prevent exploits and hacks.

game-server/src/com/aionemu/gameserver/network/aion/clientpackets/CM_USE_ITEM.java
```diff
@@ -108,23 +115,18 @@ protected void runImpl() {
...
+				case InstanceTimeClear _ -> itemAction.act(player, item, targetItem, syncId);
+				default -> itemAction.act(player, item, targetItem);
 			}
 		}
+
+		// the use delay starts when the item use is over, so actions which cast start it themselves
+		if (!player.getController().hasScheduledTask(TaskId.ITEM_USE)
+			&& !(player.isCasting() && player.getCastingSkill().getItemTemplate() != null))
+			player.startCooldown(item);
 	}
```
neon-dev, line 125: Some *Action classes have return conditions in `act()`, so we cannot always safely start the cooldown here.

game-server/src/com/aionemu/gameserver/skillengine/model/Skill.java
```diff
@@ -557,6 +559,7 @@ protected void endCast() {
 				if (!((Player) effector).getInventory().decreaseByObjectId(item.getObjectId(), 1, ItemUpdateType.DEC_ITEM_USE))
 					return;
 			}
+			((Player) effector).startCooldown(item);
```
neon-dev: GitHub doesn't let me create a code suggestion for unchanged lines above the fold, so you'll have to apply the changes manually:
```java
if (skillMethod == SkillMethod.ITEM && effector instanceof Player itemUser) {
	Item item = itemUser.getInventory().getItemByObjId(itemObjectId);
	if (item == null)
		return;
	if (item.getActivationCount() > 1) {
		item.setActivationCount(item.getActivationCount() - 1);
	} else if (!itemUser.getInventory().decreaseByObjectId(item.getObjectId(), 1, ItemUpdateType.DEC_ITEM_USE))
		return;
	}
	itemUser.startCooldown(item);
}
```

neon-dev (changes requested):

game-server/src/com/aionemu/gameserver/model/templates/item/actions/ApExtractAction.java
Birthright50: It’s possibly just client-side text for some edge cases. 1401643 does not occur anywhere in either the 4.6 or the 5.8 binary, and I can’t even think of a way to reproduce it in theory.

game-server/src/com/aionemu/gameserver/model/templates/item/actions/RideAction.java
neon-dev, line 1: Some mounts now have a lower `casting_delay` than before, but part of the mounting animation only starts playing after the old 3000ms delay.
Example mounts are 190100042 and 190100044.
Are the lower `casting_delay` values wrong for 4.8 or is there something in this class that could cause this? I couldn't find anything yet.

neon-dev: > Removed `casting_delay` support from `CompositionAction`, since retail ignores it even when present.

Maybe the same is the case for RideAction.

game-server/src/com/aionemu/gameserver/model/templates/item/actions/ExtractAction.java
```diff
@@ -58,6 +58,8 @@ public void abort() {
 		player.getController().addTask(TaskId.ITEM_USE, ThreadPoolManager.getInstance().schedule(() -> {
 			player.getObserveController().removeObserver(observer);
 			boolean result = canAct(player, parentItem, targetItem) && EnchantService.breakItem(player, targetItem, parentItem);
+			if (result)
+				player.startCooldown(parentItem);
```
neon-dev, line 64: Shouldn't the cooldown also start if the item breaks?

game-server/src/com/aionemu/gameserver/model/templates/item/actions/RideAction.java
neon-dev: As discussed via chat, this problem also exists on retail and there is nothing we can do about it. The mounting animation cannot be controlled from the server side and always plays with a fixed duration.

neon-dev (changes requested):

game-server/src/com/aionemu/gameserver/model/templates/item/actions/ExtractAction.java
Birthright50: I didn’t quite understand the question. This is specifically about when the item breaks and the cooldown starts. There’s another thing here as well: the only item with an extract action is 165000001, and it doesn’t have a use delay. So this code will almost never actually set a cooldown, it’s basically there just for consistency.

neon-dev: Never mind. I wasn't paying full attention here and got confused by the use of the EnchantService, so I thought a failure would need to be handled as well. In reality, a failure in breakItem would usually only happen from exploits.
I think it's at least worth adding a comment that there's no item that actually has a cooldown. Just so that the next time someone stumbles upon it, they don't start wondering what the seemingly non-working code is all about.
Or just remove it. I'm usually all for consistency, but not necessarily at the expense of clarity.

Birthright50: comment added

PR #156, opened 2026-07-08
FIX: update NPC look state after dialog closes.

SVDNESS:
> Fixed an issue where:
>
> * After closing an NPC dialog or moving away from the NPC, other players would still see the player as interacting with the NPC.
> * The NPC would also incorrectly appear to still be engaged in a conversation with that player.


neon-dev:
I've found another problem when there were two or more players talking to the same NPC: If you close the dialog after another player started talking to the same NPC, it would look like the NPC kept talking to you. This is because the game client doesn't update the NPC's target while you're talking to it (which means it is isolated to just your own character in this moment and needs to be addressed separately).
Sending a dedicated packet in [CM_CLOSE_DIALOG](https://github.com/beyond-aion/aion-server/pull/156/changes#diff-048c34738ce835eb0832471d2ecc6d6b96a42dbf31f0fbd55be8c2792f0932e9R35) fixes this and works even more reliably than on retail, where I could sometimes reproduce the same problem.

PR #162, opened 2026-07-31
Rework summon release handling

Birthright50:
Adjusted pet unsummon delay from 5s to 3s to match retail behavior.

Added instant pet despawn for skills with PetOrderUnSummonEffect ([Spirit's Empowerment](https://aioncodex.com/48/skill/3541/)), and for death of the pet or its master.

Reworked how release causes interact. release() bailed out whenever the pet was already in RELEASE mode, so the first cause always won and `PetOrderUnSummonEffect` did nothing while another release was pending. Each cause now carries its own delay and whether the master can cancel it, and an instant cause takes over a scheduled one.

Also fixed retail behavior where a summon cannot use any skills while being released. Skills are no longer queued during the release process.

### DIFF
--- a/game-server/src/com/aionemu/gameserver/services/summons/SummonsService.java
+++ b/game-server/src/com/aionemu/gameserver/services/summons/SummonsService.java
@@ -87,16 +87,14 @@ public class SummonsService {
 			if (summoningSkill != null && summoningSkill.getCooldown() > 0)
 				master.setSkillCoolDown(summoningSkill.getCooldownId(), summoningSkill.getCooldown() * 100 + System.currentTimeMillis());
 
-			if (unsummonType != UnsummonType.LOGOUT) { // the master is leaving the world, updating his client is pointless
-				if (unsummonType == UnsummonType.DISTANCE)
-					PacketSendUtility.sendPacket(master, SM_SYSTEM_MESSAGE.STR_SKILL_SUMMON_UNSUMMON_BY_TOO_DISTANCE());
-				else
-					PacketSendUtility.sendPacket(master, SM_SYSTEM_MESSAGE.STR_SKILL_SUMMON_UNSUMMONED(summon.getL10n()));
-				PacketSendUtility.sendPacket(master, new SM_SUMMON_PANEL_REMOVE(summon.getSummonedBySkillId()));
-				PacketSendUtility.sendPacket(master, new SM_SUMMON_OWNER_REMOVE(summon.getObjectId()));
-				if (!addedMasterHate)
-					scheduleAddMasterHate(summon);
-			}
+			if (unsummonType == UnsummonType.DISTANCE)
+				PacketSendUtility.sendPacket(master, SM_SYSTEM_MESSAGE.STR_SKILL_SUMMON_UNSUMMON_BY_TOO_DISTANCE());
+			else
+				PacketSendUtility.sendPacket(master, SM_SYSTEM_MESSAGE.STR_SKILL_SUMMON_UNSUMMONED(summon.getL10n()));
+			PacketSendUtility.sendPacket(master, new SM_SUMMON_PANEL_REMOVE(summon.getSummonedBySkillId()));
+			PacketSendUtility.sendPacket(master, new SM_SUMMON_OWNER_REMOVE(summon.getObjectId()));
+			if (!addedMasterHate)
+				scheduleAddMasterHate(summon);
 		}
 
 		public void scheduleOrRun() {
--- a/game-server/src/com/aionemu/gameserver/services/summons/SummonsService.java
+++ b/game-server/src/com/aionemu/gameserver/services/summons/SummonsService.java
@@ -83,28 +83,19 @@ public class SummonsService {
 			if (summon.equals(master.getSummon()))
 				master.setSummon(null);
 
-			switch (unsummonType) {
-				case COMMAND:
-				case DISTANCE:
-				case SUMMON_DEATH:
-				case MASTER_DEATH:
-				case UNSPECIFIED:
-				case PET_ORDER_UNSUMMON_EFFECT:
-					SkillTemplate summoningSkill = DataManager.SKILL_DATA.getSkillTemplate(summon.getSummonedBySkillId());
-					if (summoningSkill != null && summoningSkill.getCooldown() > 0)
-						master.setSkillCoolDown(summoningSkill.getCooldownId(), summoningSkill.getCooldown() * 100 + System.currentTimeMillis());
+			SkillTemplate summoningSkill = DataManager.SKILL_DATA.getSkillTemplate(summon.getSummonedBySkillId());
+			if (summoningSkill != null && summoningSkill.getCooldown() > 0)
+				master.setSkillCoolDown(summoningSkill.getCooldownId(), summoningSkill.getCooldown() * 100 + System.currentTimeMillis());
 
-					if (unsummonType == UnsummonType.DISTANCE)
-						PacketSendUtility.sendPacket(master, SM_SYSTEM_MESSAGE.STR_SKILL_SUMMON_UNSUMMON_BY_TOO_DISTANCE());
-					else
-						PacketSendUtility.sendPacket(master, SM_SYSTEM_MESSAGE.STR_SKILL_SUMMON_UNSUMMONED(summon.getL10n()));
-					PacketSendUtility.sendPacket(master, new SM_SUMMON_PANEL_REMOVE(summon.getSummonedBySkillId()));
-					PacketSendUtility.sendPacket(master, new SM_SUMMON_OWNER_REMOVE(summon.getObjectId()));
-					if (!addedMasterHate)
-						scheduleAddMasterHate(summon);
-					break;
-				case LOGOUT:
-					break;
+			if (unsummonType != UnsummonType.LOGOUT) { // the master is leaving the world, updating his client is pointless
+				if (unsummonType == UnsummonType.DISTANCE)
+					PacketSendUtility.sendPacket(master, SM_SYSTEM_MESSAGE.STR_SKILL_SUMMON_UNSUMMON_BY_TOO_DISTANCE());
+				else
+					PacketSendUtility.sendPacket(master, SM_SYSTEM_MESSAGE.STR_SKILL_SUMMON_UNSUMMONED(summon.getL10n()));
+				PacketSendUtility.sendPacket(master, new SM_SUMMON_PANEL_REMOVE(summon.getSummonedBySkillId()));
+				PacketSendUtility.sendPacket(master, new SM_SUMMON_OWNER_REMOVE(summon.getObjectId()));
+				if (!addedMasterHate)
+					scheduleAddMasterHate(summon);
 			}
 		}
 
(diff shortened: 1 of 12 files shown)


neon-dev:
> Adjusted pet unsummon delay from 5s to 3s to match retail behavior.

How did you measure the unsummon delay?
In my rudimentary tests using a stop watch, the delay seemed more like 3500 ms.

> Added instant pet despawn handling for specific skills. ([Spirit's Empowerment](https://aioncodex.com/48/skill/3541/))

Is there a video from official 4.8 servers available or did you verify this on 5.8 PTS?

Birthright50:
> How did you measure the unsummon delay?
> In my rudimentary tests using a stop watch, the delay seemed more like 3500 ms.

I performed two frame-by-frame measurements. While this method is not as accurate as packet capture, it is more than sufficient to distinguish 3 seconds from 3.5 seconds, let alone 5 seconds.
For example, with this skill: https://aioncodex.com/48/skill/3838/. On retail, using this skill does not show a summon dismissal countdown in the player's UI. Exactly 3 seconds pass between the moment the spirit uses the skill (according to the chat log) and the moment the server sends the packet indicating that the summon has been dismissed.

The same applies when the player manually dismisses the summon. The UI countdown lasts exactly 3 seconds. On retail, it disappears immediately after completing one full rotation on the action bar icons. With the current 5-second delay, however, they remain grayed out for an additional 2 seconds.
image

EDIT:
I ran a few more measurements. I'm not sure what causes it, but the results really do fluctuate between 3 and 3.5 seconds (and all of these tests were performed locally).

Frame-by-frame measurements:
7782 - 11282 (3500 ms)
10874 - 14128 (3254 ms)
1374 - 4362 (2988 ms)
720 - 3882 (3162 ms)

However, I can say with confidence that, visually, the countdown shown in the UI clearly lasts 3 seconds.

> Is there a video from official 4.8 servers available or did you verify this on 5.8 PTS?

5.8 pts

SVDNESS:
I'm more than confident that the summon logic was never changed between versions.
On Classic 5.0, the delay between the summon despawning and the summon panel disappearing is exactly **5 seconds**, not 3 seconds.
I'm curious where you got those timings from.
Here's the packet evidence.

**Summon spawned** (`C_CLIENTSIDE_NPC_ACTION`) → **Immediate click on "Dismiss Summon"** (`C_PET_ORDER`) → **Summon despawned** (`S_REMOVE_PET`):

CLIENT → SERVER
Timestamp: 08:57:32.823
Opcode:    0x1A9  (C_CLIENTSIDE_NPC_ACTION)
Length:    10
HEX
0000  A9 01 6F 56 FE 4E F7 24  80 1B                    |..oV.N.$..|
---------------------------------------

SERVER → CLIENT
Timestamp: 08:57:32.844
Opcode:    0x98  (S_ADD_PET)
Length:    41
HEX
0000  88 01 65 77 FE 4E F7 24  80 41 00 00 00 00 00 00  |..ew.N.$.A......|
0010  00 00 00 19 23 00 00 19  23 00 00 97 01 00 00 79  |....#...#......y|
0020  04 CC 05 00 00 00 00 00  00                       |.........|

SERVER → CLIENT
Timestamp: 08:57:32.844
Opcode:    0x9A  (S_CHANGE_PET_STATUS)
Length:    76
HEX
0000  8A 01 65 75 FE 41 01 00  00 00 00 00 00 00 00 00  |..eu.A..........|
0010  19 23 00 00 19 23 00 00  97 01 00 00 79 04 CC 05  |.#...#......y...|
0020  15 06 50 00 01 04 00 00  F9 04 6A 00 00 00 25 06  |..P.......j...%.|
0030  F4 17 00 00 97 01 00 00  47 03 22 05 15 06 50 00  |........G."...P.|
0040  00 00 00 00 47 04 50 00  00 00 CE 05              |....G.P.....|

CLIENT → SERVER
Timestamp: 08:57:32.844
Opcode:    0x1A8  (C_CLIENTSIDE_NPC_MOVE)
Length:    23
HEX
0000  A8 01 6F 57 FE 4E F7 24  80 A2 70 A4 44 33 DC A4  |..oW.N.$..p.D3..|
0010  44 D5 0C D3 42 2C 00                              |D...B,.|

CLIENT → SERVER
Timestamp: 08:57:33.628
Opcode:    0x13B  (C_PET_ORDER)
Length:    18
HEX
0000  3B 01 6F C4 FE 03 00 00  00 00 00 00 00 00 00 00  |;.o.............|
0010  00 00                                             |..|
---------------------------------------

SERVER → CLIENT
Timestamp: 08:57:33.649
Opcode:    0x19  (S_MESSAGE_CODE)
Length:    25
HEX
0000  09 00 65 F6 FF 19 00 00  00 00 00 8B 4F 12 00 01  |..e.........O...|
0010  24 00 FB AE 0A 00 00 00  00                       |$........|

SERVER → CLIENT
Timestamp: 08:57:33.712
Opcode:    0x9A  (S_CHANGE_PET_STATUS)
Length:    76
HEX
0000  8A 01 65 75 FE 41 03 00  00 00 00 00 00 00 00 00  |..eu.A..........|
0010  19 23 00 00 19 23 00 00  97 01 00 00 79 04 CC 05  |.#...#......y...|
0020  15 06 50 00 01 04 00 00  F9 04 6A 00 00 00 25 06  |..P.......j...%.|
0030  F4 17 00 00 97 01 00 00  47 03 22 05 15 06 50 00  |........G."...P.|
0040  00 00 00 00 47 04 50 00  00 00 CE 05              |....G.P.....|

SERVER → CLIENT
Timestamp: 08:57:38.872
Opcode:    0x16  (S_REMOVE_OBJECT)
Length:    10
HEX
0000  0E 00 65 F1 FF 4E F7 24  80 01                    |..e..N.$..|

SERVER → CLIENT
Timestamp: 08:57:38.872
Opcode:    0x19  (S_MESSAGE_CODE)
Length:    25
HEX
0000  09 00 65 F6 FF 19 00 00  00 00 00 86 4F 12 00 01  |..e.........O...|
0010  24 00 FB AE 0A 00 00 00  00                       |$........|

SERVER → CLIENT
Timestamp: 08:57:38.872
Opcode:    0x99  (S_REMOVE_PET)
Length:    9
HEX
0000  89 01 65 76 FE 4E F7 24  80                       |..ev.N.$.|
---------------------------------------

As you can see, exactly **5 seconds** pass between pressing the **"Dismiss Summon"** button and the server sending `S_REMOVE_PET`.

I'm not aware of Retail 4.8 behaving any differently.

The dismissal of Elementals via skills such as **Spirit's Empowerment** has its own, separate despawn timing system. [[Spirit's Empowerment](https://aioncodex.com/us/skill/3542/)]

In other words, it follows different timing than the standard summon dismissal.

SVDNESS:
Also, on Retail, the behavior has always been as follows:

* If the player teleports within the same world, the summon is **not** despawned. Instead, it is teleported to its owner's new position.
* If the player teleports to a different world, the Elemental is **not** dismissed before the world transfer. Instead, it is dismissed **after** the new world has finished loading.

If this behavior is needed, I can prepare a separate PR for it.

Birthright50:
> I'm more than confident that the summon logic was never changed between versions.
> On Classic 5.0, the delay between the summon despawning and the summon panel disappearing is exactly **5 seconds**, not 3 seconds.
> I'm curious where you got those timings from.

You shouldn't test this functionality on Classic, and especially not on version 5.0. And yes, the reason is that in those versions the unsummon delay really is 5 seconds, it's even clearly noticeable from the UI countdown (which is obviously synced with the backend, since the backend doesn't send any countdown values at all). Just out of curiosity I logged into classic myself and got about 5.5s between the two log messages, the same ~500ms jitter I mentioned above when testing on 4.6 and 5.8. You need to test on the versions that are close to 4.8: 4.6 and 5.8 - there it's 3 seconds.
As for the "Spirit's Empowerment" comment, I don't really follow it - the only difference there is that the summon gets unsummoned instantly.

Birthright50:
> * If the player teleports to a different world, the Elemental is **not** dismissed before the world transfer. Instead, it is dismissed **after** the new world has finished loading.

Again, that matches Classic, but not 4.8 - here the summon shouldn't disappear even when moving between locations. This mechanic was introduced starting with 4.6: if the summon gets stuck while in Guard state and doesn't actually move, and the distance to its master breaks, the summon automatically teleports to its owner. Also, moving within the same location or between locations doesn't trigger a distance-based unsummon.

SVDNESS:
I agree that testing against a version closer to 4.8 makes more sense.
That said, I'm not convinced that the summon mechanics themselves changed between versions.
I don't think 4.6 PTS is a reliable reference either, since many gameplay mechanics were changed between 4.6 and 4.8. The same is true for skills—there were significant changes there as well. Because of that, using 4.6 PTS as the baseline isn't ideal.
The same applies to 5.8. A lot of things were changed by the developers, from the starting areas to various gameplay mechanics.
The most reliable way to settle this is to reverse engineer `game.dll` and verify the behavior directly. That's what I'm going to do.

Birthright50:
> I agree that testing against a version closer to 4.8 makes more sense. That said, I'm not convinced that the summon mechanics themselves changed between versions. I don't think 4.6 PTS is a reliable reference either, since many gameplay mechanics were changed between 4.6 and 4.8. The same is true for skills—there were significant changes there as well. Because of that, using 4.6 PTS as the baseline isn't ideal. The same applies to 5.8. A lot of things were changed by the developers, from the starting areas to various gameplay mechanics. The most reliable way to settle this is to reverse engineer `game.dll` and verify the behavior directly. That's what I'm going to do.

I'll say it again: server timings need to be synced with the UI - the UI timer on the 4.6, 4.8, and 5.8 clients is set to 3 seconds. On current Classic it's 5 seconds. There's no point citing Classic, especially in its current state.

Birthright50:
https://github.com/beyond-aion/aion-patch-notes/blob/main/Gameforge%20(EU)/3.0%20-%202012-08-14/PatchNote_3_0_EN.pdf

> Decreased the time it takes to dismiss a summoned spirit from 5 seconds to 3 seconds.

SVDNESS:
Turns out I didn't even have to reverse engineer `game.dll`.
I found a 10-year-old official 4.8 NA gameplay video and verified the behavior.
You were right: after pressing the **Dismiss Summon** button, the summon is despawned exactly **3 seconds** later, and the summon panel disappears at the same time.
I'm honestly surprised the developers changed this behavior in Classic. I was convinced this particular mechanic had remained unchanged across versions.
Thanks for the discussion!

Birthright50:
> Turns out I didn't even have to reverse engineer `game.dll`.
> I found a 10-year-old official 4.8 NA gameplay video and verified the behavior.
> You were right: after pressing the **Dismiss Summon** button, the summon is despawned exactly **3 seconds** later, and the summon panel disappears at the same time.
> I'm honestly surprised the developers changed this behavior in Classic. I was convinced this particular mechanic had remained unchanged across versions.
> Thanks for the discussion!

more likely сlassic's developers tried to preserve all the mechanics that were in the vanilla 1.–2. versions (specifically, in the vanilla versions it was indeed 5 seconds before version 3.0, I posted the vanilla 3.0 changelog above), so they didn't change it in 3.0 and beyond either, that's exactly where сlassic starts to diverge significantly from vanilla Aion

game-server/src/com/aionemu/gameserver/services/summons/SummonsService.java
```diff
@@ -83,28 +83,19 @@ public void run() {
...
-					if (!addedMasterHate)
-						scheduleAddMasterHate(summon);
-					break;
-				case LOGOUT:
-					break;
+			SkillTemplate summoningSkill = DataManager.SKILL_DATA.getSkillTemplate(summon.getSummonedBySkillId());
+			if (summoningSkill != null && summoningSkill.getCooldown() > 0)
+				master.setSkillCoolDown(summoningSkill.getCooldownId(), summoningSkill.getCooldown() * 100 + System.currentTimeMillis());
+
+			if (unsummonType != UnsummonType.LOGOUT) { // the master is leaving the world, updating his client is pointless
```
neon-dev: This check can be removed to keep things simple.
The PacketSendUtility does not send any packets to players who are logging out and scheduleAddMasterHate also checks for isOnline already.

PR #169, opened 2026-08-16
Fix commands crashing on missing parameters

Birthright50:
10 commands read parameters they never checked for, throwing ArrayIndexOutOfBoundsException (logged server-side, no feedback to the sender):
`///wishid`
`///set_enchantcount`
`///attrbonus`
 `///combineskill`,
`//damage`
`//rift open`
`//weather set`
`//passkeyreset`
 `//stoken show`
`//remove`.


neon-dev:
These are all commands that need to be reworked or at least updated. Most commands that we haven't touched yet contain countless more problems. They are also not sending proper syntax info yet (if at all) and often send wrong or cryptic messages to the user for certain actions.
The legacy `info` methods are also obsolete. Reworked commands use the provided `sendInfo` method instead.
The superclasses AdminCommand and ConsoleCommand still have an old one arg constructor with a TODO comment saying these should be removed once all commands using them have been ported to the new two arg constructor.

If you're interested in reworking some commands, check one of the commands that already use the two arg constructor, like `MoveTo`. It shows how to format the syntax info so the highlighting works correctly. It is also not required anymore to catch unchecked exceptions like NumberFormatException or other instances of IllegalArgumentException. A useful error message is sent to the user automatically if they enter an invalid value that causes such an exception.

So I will likely only merge the commonly used commands in this PR. This is because I see very little point in merging partial fixes for something that should likely be thrown away completely. From my experience, some commands are not worth to be fixed. Rewriting them is often the better option.

Birthright50:
> These are all commands that need to be reworked or at least updated. Most commands that we haven't touched yet contain countless more problems. They are also not sending proper syntax info yet (if at all) and often send wrong or cryptic messages to the user for certain actions. The legacy `info` methods are also obsolete. Reworked commands use the provided `sendInfo` method instead. The superclasses AdminCommand and ConsoleCommand still have an old one arg constructor with a TODO comment saying these should be removed once all commands using them have been ported to the new two arg constructor.
> 
> If you're interested in reworking some commands, check one of the commands that already use the two arg constructor, like `MoveTo`. It shows how to format the syntax info so the highlighting works correctly. It is also not required anymore to catch unchecked exceptions like NumberFormatException or other instances of IllegalArgumentException. A useful error message is sent to the user automatically if they enter an invalid value that causes such an exception.
> 
> So I will likely only merge the commonly used commands in this PR. This is because I see very little point in merging partial fixes for something that should likely be thrown away completely. From my experience, some commands are not worth to be fixed. Rewriting them is often the better option.

Maybe, but it shouldn’t really hurt if everything is merged together at once.

PR #178, opened 2026-08-27
Name and type the transform restriction flags

Birthright50:
The seven restriction flags of shapechange, polymorph and deform were ints, five of them named like in `skill_base.xml` or `client_skill_base.xml` **`(reserved*)`**, which said nothing about what they do. They are now booleans named cantRecall (res1), cantJump (res2), cantAttack (res3), cantUseItems (res5), cantFly (res6), cantUseSkills and cantMove, written as true like every other boolean attribute in the file.

Every one of the 746 values in our data was 1, except skill 17012 Petrification, which carries 100 in the client too and therefore never matched the `== 1` comparison the code used. Retail treats the flag as set for any value above zero, verified in 5.8 pts.


game-server/src/com/aionemu/gameserver/network/aion/serverpackets/SM_TRANSFORM.java
```diff
@@ -69,14 +69,14 @@ protected void writeImpl(AionConnection con) {
...
-			writeC(creature.getTransformModel().getBanUseSkills());
+			writeC(creature.getTransformModel().cantUseSkills() ? 1 : 0);
 			writeD(creature.getTransformModel().getType().getId());
-			writeC(creature.getTransformModel().getRes6());
-			writeC(creature.getTransformModel().getRes5());
-			writeC(creature.getTransformModel().getRes3());
-			writeC(creature.getTransformModel().getRes2());
-			writeC(creature.getTransformModel().getRes1());
-			writeC(creature.getTransformModel().getBanMovement());
+			writeC(creature.getTransformModel().cantFly() ? 1 : 0);
```
neon-dev: Have you checked if it makes a difference that 100 is no longer sent here for Petrification?
I am wondering if res6 being 100 for just a single skill had any significance.

game-server/src/com/aionemu/gameserver/network/aion/serverpackets/SM_TRANSFORM.java
```diff
@@ -42,8 +42,8 @@ public SM_TRANSFORM(Creature creature, int modelId, int unk7, TransformType type
 
 	/**
 	 * structure SM_TRANSFORM D - objectId D - modelId (res9) 100% H - state F - 0,25f F - 2.0f C - cannotuseskill(res7) 100%- not used for FORM1 D -
-	 * transformTypeId (res8) 100% C - (res6) cannot fly/glide? C - cannot useitem (res5) C - attack disabled (res3) C - jump disabled (res2?) C -
-	 * summon disabled? (res1?) C - move disabled(res13) 100% D - panelId(res4) 100% cant mount - client recognises it by modelId? TODO server side
+	 * transformTypeId (res8) 100% C - (res6) cannot fly/glide C - cannot useitem (res5) C - attack disabled (res3) C - jump disabled (res2) C -
+	 * summon disabled (res1) C - move disabled(res13) 100% D - panelId(res4) 100% cant mount - client recognises it by modelId? TODO server side
 	 * checks for mounts
 	 */
```
neon-dev: This can be removed. The TODO comment could be kept, but I believe it's obsolete as well.

game-server/src/com/aionemu/gameserver/network/aion/serverpackets/SM_TRANSFORM.java
Birthright50: It didn’t make any difference.

game-server/src/com/aionemu/gameserver/network/aion/serverpackets/SM_TRANSFORM.java
Birthright50: The whole comment?

neon-dev: Well yeah, unless you see some info in there that is not redundant or obsolete.

PR #180, opened 2026-08-29
Interrupt casts on hide only for those who can no longer see the target.

SVDNESS:
### Problem

HideEffect cancels every cast aimed at the hiding creature, even though the comment above the call says it is meant for enemy casts only. Group members keep seeing a hidden ally, since Player.canSee returns true for members of the same team. As a result, a healer casting on an assassin loses the cast the moment the assassin uses Hide, even though the target remains visible and the line of sight is still intact.

### Solution

The next lines of the same method already establish the correct behavior: removeTargetFrom(effected, true) clears the target only for players who cannot see the creature.

cancelCastOn now uses the same condition, so a cast is interrupted exactly when its target becomes invisible to the caster: a hidden group member keeps being healed, an enemy loses the cast as before, and an enemy with a high enough perception level to still detect the target keeps casting, which matches retail behavior.

The target comparison was also changed to target.equals(getFirstTarget()), because getFirstTarget() can be null for point-targeted skills and would otherwise cause an exception.

### Testing

Tested in game with two clients:

* A heal on a grouped assassin who uses Hide → the cast completes.
* The same heal on an assassin outside the group → the cast is interrupted.
* An attack skill on an enemy assassin → the cast is interrupted as before.

### DIFF
--- a/game-server/src/com/aionemu/gameserver/controllers/attack/AttackUtil.java
+++ b/game-server/src/com/aionemu/gameserver/controllers/attack/AttackUtil.java
@@ -546,11 +546,13 @@ public class AttackUtil {
 		return AttackStatus.NORMALHIT;
 	}
 
+	//Cast is interrupted only for those who can no longer see the target after it enters hide.
 	public static void cancelCastOn(Creature target) {
 		target.getKnownList().forEachObject(visibleObject -> {
 			if (visibleObject instanceof Creature creature && visibleObject.getTarget() == target) {
-				if (creature.getCastingSkill() != null && creature.getCastingSkill().getFirstTarget().equals(target))
+				if (creature.getCastingSkill() != null && target.equals(creature.getCastingSkill().getFirstTarget()) && !creature.canSee(target)) {
 					creature.getController().cancelCurrentSkill(null);
+				}
 			}
 		});
 	}


Birthright50:
On retail, the cast isn’t interrupted when the `hide` effect is applied. The checks are only performed at the end of the cast.
So I think it would be better to remove this line from [HideEffect.java:64](https://github.com/beyond-aion/aion-server/blob/1b0e2210522f1988b8115d2222b5f58596968600/game-server/src/com/aionemu/gameserver/skillengine/effect/HideEffect.java#L64) and add the check to [Skill.java:226](https://github.com/beyond-aion/aion-server/blob/1b0e2210522f1988b8115d2222b5f58596968600/game-server/src/com/aionemu/gameserver/skillengine/model/Skill.java#L226):
```java
if (!player.canSee(target))
    return false;
```

game-server/data/handlers/ai/GeneralNpcAI.java
```diff
@@ -41,6 +41,8 @@ protected boolean handleCreatureNeedsSupport(Creature creature) {
 	@Override
 	protected void handleCreatureNotSee(Creature creature) {
 		if (creature.equals(getTarget())) {
+			if (getOwner().isCasting() && !getOwner().canSee(creature))
+				return; // the cast runs out and simply misses the hidden target
 			getOwner().getController().abortCast();
 			onGeneralEvent(AIEventType.TARGET_TOOFAR);
 		}
```
neon-dev: Isn't `handleCreatureNotSee()` already tied to the same `canSee()` check?
If so, I think we can simplify it all to this:
```suggestion
		if (creature.equals(getTarget()) && !getOwner().isCasting()) // the cast runs out and simply misses the hidden target
			onGeneralEvent(AIEventType.TARGET_TOOFAR);
```

neon-dev: It's also important to check similar cases other than hide, like a despawning mob for example.

Birthright50: > Isn't `handleCreatureNotSee()` already tied to the same `canSee()` check? If so, I think we can simplify it all to this:

`notSee` also fires from `KnownList.del` (teleport, despawn), so a mob would still hit a player who teleported away mid cast.
Applied your simplification (keeping `abortCast()` for a skill that is not started yet) and moved the target checks to the end of the cast, which also replaces the death observer, so a dying target no longer cuts a cast short.

game-server/src/com/aionemu/gameserver/skillengine/model/Skill.java
```diff
@@ -201,10 +194,23 @@ private boolean canPayCastCosts() {
...
 	private boolean validateEffectedList() {
 		if (effector instanceof Player player) {
-			if (canUseSkill(player))
+			if (canUseSkill(player)) {
+				if (!canTargetFirstTarget()) {
+					if (getTargetRangeAttribute() != TargetRangeAttribute.AREA) {
+						PacketSendUtility.sendPacket(player, SM_SYSTEM_MESSAGE.STR_SKILL_TARGET_IS_NOT_VALID());
+						return false;
+					}
+					effectedList.remove(firstTarget); // everyone else in the area is still hit
```
neon-dev, line 203: Shouldn't `AREA` and `POINT` skills still be able to hit the target?

Birthright50 (dismissed):

PR #182, opened 2026-08-29
SM_INVINCIBLE_TIME: implement retail packet.

SVDNESS:
## Summary

Implemented the `SM_INVINCIBLE_TIME` packet used when the player enters and leaves the `BLINKING` state.

## Changes

* Send `SM_INVINCIBLE_TIME(60_000)` when the protection task starts and `BLINKING` is applied.
* Send `SM_INVINCIBLE_TIME(0)` when the protection task ends and `BLINKING` is removed.
* The packet is sent to the player directly, while `SM_PLAYER_STATE` is broadcast to nearby players to update the visual state.

## Behavior

The packet keeps the client synchronized with the server-side protection timer and informs it when the temporary invincibility period starts and ends.

### DIFF
--- a/game-server/src/com/aionemu/gameserver/controllers/PlayerController.java
+++ b/game-server/src/com/aionemu/gameserver/controllers/PlayerController.java
@@ -646,6 +646,7 @@ public class PlayerController extends CreatureController<Player> {
 			AttackUtil.cancelCastOn(getOwner());
 			AttackUtil.removeTargetFrom(getOwner());
 			PacketSendUtility.broadcastToSightedPlayers(getOwner(), new SM_PLAYER_STATE(getOwner()), true);
+			PacketSendUtility.sendPacket(getOwner(), new SM_INVINCIBLE_TIME(60_000));
 			addTask(TaskId.PROTECTION_ACTIVE, ThreadPoolManager.getInstance().schedule(this::stopProtectionActiveTask, 60000));
 		}
 	}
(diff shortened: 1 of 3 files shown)


game-server/src/com/aionemu/gameserver/controllers/PlayerController.java
```diff
@@ -646,6 +646,7 @@ public void startProtectionActiveTask() {
 			AttackUtil.cancelCastOn(getOwner());
 			AttackUtil.removeTargetFrom(getOwner());
 			PacketSendUtility.broadcastToSightedPlayers(getOwner(), new SM_PLAYER_STATE(getOwner()), true);
+			PacketSendUtility.sendPacket(getOwner(), new SM_INVINCIBLE_TIME(60_000));
```
Birthright50, line 649: The main issue is that `invincible` isn’t extended. For example, if you teleport to any loc and don’t move (so the character is blinking), after 30 seconds you get teleported again using `RecallInstant`. In that case, the `invincible` state should be extended for another 60 seconds, but the timer continues from the initial `invincible` activation.

I think it would be better to handle it like this:

```java
private static final int PROTECTION_TIME = 60000;

public void startProtectionActiveTask() {
    Player player = getOwner();
    if (!player.isProtectionActive()) {
        player.setVisualState(CreatureVisualState.BLINKING);
        AttackUtil.cancelCastOn(player);
        AttackUtil.removeTargetFrom(player);
        PacketSendUtility.broadcastToSightedPlayers(player, new SM_PLAYER_STATE(player), true);
    }
    addTask(TaskId.PROTECTION_ACTIVE, ThreadPoolManager.getInstance().schedule(this::stopProtectionActiveTask, PROTECTION_TIME));
    PacketSendUtility.sendPacket(player, new SM_INVINCIBLE_TIME(PROTECTION_TIME));
}

Birthright50 (dismissed):

PR #183, opened 2026-08-30
Rework Summon Group Member

Birthright50:
Follows up #177, which implemented the effect. This replaces the checks and the request handling that do not match retail.

The main reference point for testing is always the same: whether the recall icon is "on" on the client side. In prisons it is "on", meaning you can recall there. All of these restrictions apply exclusively to the caster. (For testing I used Shift+F1 and cast the skill from there, this restriction can actually be bypassed on the client.)

As for the priority, I can roughly explain how it works (client_world.xml):
```xml
<recallareas>
    <ispossible>TRUE</ispossible>         <!-- 1 -->
    <recallarea>                          <!-- 2 -->
        <name>LC1_ReCallArea_1</name>
        <possible>FALSE</possible>
        <priority>10</priority>
        <points_info><type>polygon</type><points>...</points></points_info>
    </recallarea>
</recallareas>

<limitareas>
    <limitarea>                           <!-- 3 -->
        <name>Limit_Area_1</name>
        <norecall>FALSE</norecall>
        <noride>...</noride><nofly>...</nofly>
        <priority>16</priority>
        <points_info>...</points_info>
    </limitarea>
</limitareas>
```
1) The default for the entire map. In Sanctum it's true, while in Tiamaranta it's false.
2) These are polygons specifically for recall. possible=false means recall is not allowed here. In Sanctum or Pandaemonium, for example, this is the entrance to the Secret Library.
3) These are universal restriction polygons. A single polygon can specify restrictions for recall (norecall), riding (noride) and flying all at once. norecall=FALSE means recall is not prohibited here, so it is allowed.

The priority matters when multiple polygons overlap on the map, for example when a smaller polygon is located inside a larger one. The polygon with the lower priority number takes precedence over the one with the higher number. Without priorities it's basically a lottery: whichever polygon happens to be checked first determines which rule is applied.

The complete algorithm:
1) Get all zones the caster is physically standing inside (`findZones()`).
2) Discard everything that isn't LIMIT. The flags for SUB, PVP, FLY and ITEM_USE zones have nothing to do with recall.
3) Discard zones with flags="-1", this means no data, not everything is prohibited.
4) From the remaining zones, select the one with the lowest priority.
5) If it doesn't have the RECALL bit then prohibited, done.
6) Otherwise check the map: does it have the RECALL flag in world_maps.xml? No -> prohibited. Yes -> allowed.

For step 4 to work, the priority attribute of all 357 LIMIT zones is now filled from the client world data.

### Pending requests

Pending requests now live in `RecallService`, which holds the destination, times out after 30 seconds and cancels the request whenever the summoned player enters combat, dies, gathers, crafts, opens a private store, polymorphs into another tribe, gets teleported or logs out. Retail never revalidates on accept, it cancels the pending request instead, so the teleport on accept stays unconditional.

The generic response requester cannot express that, since it only knows how to answer, and the client closes the summon window on its own only when it answered, so `SM_RECALLED_BY_OTHER` now also closes it (first byte 1, no name).

The caster is no longer refused while in combat and is refused while flying or gliding, which it does.


game-server/src/com/aionemu/gameserver/services/player/PlayerLeaveWorldService.java
```diff
@@ -77,6 +78,7 @@ public static void leaveWorld(Player player) {
 		}
 
 		FindGroupService.getInstance().onLogout(player);
+		RecallService.getInstance().cancel(player, CancelReason.SILENT);
```
neon-dev: Why not use `CancelReason.CANCELLED` so that the caster gets notified?
Even if retail doesn't do this, it seems like an oversight by the devs rather than a deliberate choice.
We could also completely remove the enum value `CancelReason.SILENT` that way.

Birthright50: Done. Retail has no cancel hook on logout at all: the pending request lives in fields on the User object and simply dies with the session, so nothing is sent because there is nothing left to send it from, not because it was a decision. And `SILENT `was worse than that anyway, since we drop the request the caster never even got the 30sec timeout message. `SILENT `is gone.

PR #185, opened 2026-08-31
Cancel pending summon requests on all teleport paths

SVDNESS:
Follow-up to b700b44 (Rework Summon Group Member), based on the packet captures I posted in the commit comments and on in-game testing of changes on a your 4.8 server.

### The cancel hook misses most teleport paths

`RecallService.cancel` is only called from the full `teleportTo(player, worldId, instanceId, x, y, z, heading, animation)` overload, but that is not the funnel — several paths reach the world change without passing through it:

- `teleport(player, location, animation)` calls `sendLoc(...)` directly in the instant branch, so **NPC teleporters bypass the hook entirely**
- the same method's `FLIGHT` branch never touches it either
- `teleportTo(player, WorldPosition)` goes through `spawnOnSameMap(...)` when the target map is the current one
- `teleportDeadTo(...)`

Symptom in game: the summoned player teleports away with the window open, the request stays pending, and the caster only learns about it 30 seconds later from the server timeout. The client's countdown also goes visibly wrong once the window survives a map change — it showed 45 s, 90 s and 75 s across three runs instead of the remaining time.

Moved the hook into `sendLoc(...)`, which is the actual common point for every instant teleport, and added explicit calls to the same-map branch, `teleportDeadTo(...)` and the `FLIGHT` branch. Removed it from `teleportTo(...)`, where it is now redundant.

### Testing

Tested in game on 4.8, both changes together, with the window open in each case:
- target teleports via NPC teleporter, return scroll, bind point, within the same map, and by taking a flight path — window closes immediately, caster gets `DONOT_ACCEPT`

### DIFF
--- a/game-server/src/com/aionemu/gameserver/network/aion/serverpackets/SM_RECALLED_BY_OTHER.java
+++ b/game-server/src/com/aionemu/gameserver/network/aion/serverpackets/SM_RECALLED_BY_OTHER.java
@@ -10,27 +10,24 @@ import com.aionemu.gameserver.network.aion.AionServerPacket;
  * @author SVDNESS
  */
 public class SM_RECALLED_BY_OTHER extends AionServerPacket {
-
+	
+	private static final int OPEN_WINDOW = 0;
+	private static final int CANCELLED = 1;
+	private final int result;
 	private final String casterName;
 	private final int skillId;
 	private final int seconds;
-
-	/**
-	 * Closes the window on the client.
-	 */
-	public SM_RECALLED_BY_OTHER() {
-		this(null, 0, 0);
+	
+	public static SM_RECALLED_BY_OTHER cancelled() {
+		return new SM_RECALLED_BY_OTHER(CANCELLED, null, 0, 0);
 	}
 
-	/**
-	 * @param casterName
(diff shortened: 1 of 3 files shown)


neon-dev:
Thank you for testing and submitting further improvements.
I have reverted the changes to the unrelated classes, as introducing a static factory method for only one constructor was inconsistent and I don't think we need to reinvent the wheel here. Let's just focus on the fixes.

More importantly, I have pushed a commit fixing several more problems:
- Removed redundant recall cancellation in `teleportDeadTo()`. `Player.onDie()` already cancels it and dead players cannot receive a summon request
- Fixed the recall request would not be cancelled if received during the teleport animation (moved into `abortActions()`, which is called from `SpawnTask.run()` again in such cases)
- Fixed the recall cancellation when starting a flight teleport only covered the `SecurityConfig.ENABLE_FLYPATH_VALIDATOR` path
- `abortActions()` is now also called when starting a flight teleport for additional safety against abuse and invalid player states
- Fixed teleporting on the same map didn't cancel the ride action, causing an invalid player state on the client side.

SVDNESS:
Np, thx u 👍

PR #189, opened 2026-09-02
Item actions audit

Birthright50:
Follows #153, #184.

* Align item use and equip validation with retail checks, order and messages.
* Fix missing equip restrictions and derive mastery requirements from mastery effects.
* Add persistent 10-minute grace periods for lost abyss rank restrictions (require database migration).
* Restrict soul bind to valid player states.
* Fix the [item cast cancel on target switch](https://github.com/beyond-aion/aion-server/commit/5de52910fe367b6ad3320d68dc5e31cebe792b7d) regression. (see comment below)
* Fix gathering observer handling and add the missing cancellation cases.
* Fix `SM_ITEM_USAGE_ANIMATION `packet fields (use enum instead of magic numbers and clarify unk fields).
* Use the `casting_delay `from the skill for items with `activation_mode `set to `Combat`.


SVDNESS:
Nice PR. I am considering taking part of this implementation, so while having tea I went through it, and there are a few interesting points worth a second look.

**1.** `CreatureState.getActionState()` reports `LOOTING` for a player on a flying mount. `RideAction` sets `RESTING` and additionally `FLOATING_CORPSE` when `isInFlyingState()`, so the resulting state is `0xC` — the same value the `switch` maps to `LOOTING`.

The block itself is correct (`isStanding` is false either way); only the message parameter is wrong. The state value alone cannot disambiguate these two cases, so `PlayerMode.RIDE` would have to be checked at the call site before falling back to `getActionState()`.

**2.** `ItemGroup`: the `ALL_ARMOR` groups declare `requiresMastery = true`, but nothing can ever satisfy that requirement. There is no `ArmorMasteryEffect` with `armor="ALL_ARMOR"`, so `getMasterySkills()` returns an empty set and `checkAvailableEquipSkills()` silently passes.

That covers `TORSO`, `GLOVE`, `SHOES`, `PANTS` and `SHOULDER` — in the item data I looked at, roughly 6,000+ items, which previously required skill `103/106`.

The end result happens to be fine, since `103 "Basic Cloth Armor Proficiency"` is auto-learned at level 1 by every base class, so the check was effectively a no-op anyway. But it now passes by accident rather than by declaration. Either set those five groups to `false`, or map `ALL_ARMOR` to the `ROBE` mastery.

**3.** `scheduleRankLimitCheck()`: `secondsLeft > RANK_LIMIT_LAST_WARNING_SECONDS` skips the 1-minute warning when the remaining time is exactly 60 seconds.

Changing it to `>=` alone does not work either, because `warnAboutRankLimitItems()` calls back into `scheduleRankLimitCheck(60)` and the warning would then repeat forever. Having the warning method schedule the unequip directly fixes both.

**4.** `SM_ITEM_USAGE_ANIMATION` changes the wire format for almost every call, and that is not mentioned in the commit message. I checked it against a capture from a PTS, and the interpretation is correct — so this is mostly a request to document it.

The body layout matches the new code exactly (28 bytes): `D playerObjId, D targetObjId, D itemObjId, D itemId, D castTime, C stage, C suppress, H count, D element`. Five packets captured: an HP/MP potion, buff food, a teleport scroll (START and SUCCESS) and an emotion book.

* The two bytes after `stage` are `01 00` in every single packet, on both the START and the SUCCESS stage → `count = 1`. The new reading is right, and the old default (`unk1 = 0, unk2 = 1`, i.e. `00 01` = 256 on the wire) was wrong.
* `castTime` is confirmed by data: the teleport scroll START carried `10 27 00 00` = 10000 ms, and the SUCCESS packet followed 10.001 s later.
* `stage` is `0` on START and `1` on SUCCESS, `suppress` is `0` in all of them.
* The trailing `D` is definitely **not** a cooldown: a potion on a 30-second cooldown and a buff food on a 1-hour cooldown both carried `256`; the teleport scroll carried `0` on both stages; the emotion book carried `0x9DC65A4A`, which looks like uninitialised buffer content. Four different values with identical client behaviour, so "counted but never read" holds and hardcoding `0` is safe.

The part worth adding to the commit message is the two silent side effects: pet feeding used to put `15360` into that trailing `D`, and the three-argument constructor used to put `1` — both are now `0`. The old code even carried a `// mb cd?` comment there, so it deserves an explicit line.

**5.** `ItemUseAnimation` naming and documentation.

The javadoc says each group is a `START / SUCCESS / CANCEL` triple, but the `USE` group has four values, and `5` and `7` are unexplained gaps.

Also `REIDENTIFY_CANCEL(14)`: `TuningAction` sends `14` from `abort()` and again from the "target item is gone" failure path. By analogy with the `USE` group, where `2 = FAIL` and `3 = CANCEL`, `14` may actually be the failure stage, with the cancel stage at `15`.

**6.** `GATHERING_CANCEL` has no `DEATH` bit. Correct as far as the observer goes, since it does not override `died()` — but is gathering cancelled on death somewhere else?

**7.** In `canUseItem()` the comment that justified the previous ordering was removed:

> `"Checked before the 'no actions' fallback below so a race mismatch reports correctly even without one"`

and that exact check was then moved. It would be good to state whether the new order comes from a sniff or from reasoning. The same question applies to hoisting the rank and mastery checks to the top of `equipItem()`.

**8.** `ExtractAction`: stray space in `parentItem.getItemTemplate() .getTemplateId()`.

**9.** Minor: `CM_GATHER.runImpl()` now dereferences `player` unconditionally at the top of the method. If `getActivePlayer()` is guaranteed non-null for this packet, ignore this one.

**10.** `SkillData.indexMasterySkills()` can kill the whole data load. `masterySkillsByWeapon` and `masterySkillsByArmor` are `EnumMap`s, and `EnumMap.computeIfAbsent(null, …)` throws an NPE. In the schema I looked at, `weapon` on `WeaponMasteryEffect` and `armor` on `ArmorMasteryEffect` are not marked `use="required"`, so a single skill entry missing that attribute would abort `afterUnmarshal()` and take the server down at startup rather than degrading gracefully. Current data happens to have all of them filled in, so it does not fire today. Skipping such effects with a warning would be safer:
```java
private void indexMasterySkills(int skillId, List<EffectTemplate> effects) {
    for (EffectTemplate effect : effects) {
        switch (effect) {
            case WeaponMasteryEffect e when e.getItemGroup() == null -> logMissingMasteryTarget(skillId, "weapon");
            case ArmorMasteryEffect e when e.getArmorType() == null -> logMissingMasteryTarget(skillId, "armor");
            case WeaponMasteryEffect e -> masterySkillsByWeapon.computeIfAbsent(e.getItemGroup(), _ -> new HashSet<>()).add(skillId);
            case ArmorMasteryEffect e -> masterySkillsByArmor.computeIfAbsent(e.getArmorType(), _ -> new HashSet<>()).add(skillId);
            case ShieldMasteryEffect _ -> shieldMasterySkills.add(skillId);
            default -> {}
        }
    }
}

private void logMissingMasteryTarget(int skillId, String attribute) {
    LoggerFactory.getLogger(SkillData.class).warn("Mastery effect of skill {} has no {} attribute, it grants no equip right.", skillId, attribute);
}
```


### What I liked, with evidence

Deriving mastery from the effects is strictly better than the hardcoded arrays. In the skill data I checked it additionally picks up stigma skills the arrays missed:

* Stigma Sword / Mace / Dagger / Staff / Spellbook / Shield Training — `11474`, `11477`, `11479`, `11480`/`11486`, `11481`/`11487`, `11482`
* Wear Stigma Chain / Leather Armor — `11473`, `11476`/`11485`
* Wear Plate Armor — `11499`

which means a class holding the matching armor stigma can finally equip that armor.

### It also closes a hole in KEYBLADE
The group listed `{112, 115}`. `112` is `Basic Pistol Training I`, a `weapon="GUN"` effect that every `ENGINEER` auto-learns at level 1, so since the check is an OR, any gunner passed the cipher-blade mastery check as well. Every other group in that table lists the basic and advanced skill of its own weapon, and there is no advanced cipher-blade skill at all, so `112` looks like a copy-paste slip from the `GUN` row. Deriving from the effects leaves `KEYBLADE = {115}`, the `RIDER`'s own skill, and riders keep their guns through the `GUN` set, which still contains `112`.

SVDNESS:
One case is left in `checkRankLimitItems()`. It iterates a snapshot of `getEquippedItems()`, and `unEquipItem()` takes the off-hand weapon off along with the main hand. With the new `setRankLimitExpireTime(0)` in `unEquip()`, the off-hand is still in that snapshot when the loop reaches it, now with a zeroed deadline, so it is treated as a fresh violation: the player gets a second `STR_MSG_UNEQUIP_RANKITEM_TIMER_10M` for an item that is already back in his bag, and a stale deadline is persisted on it. `MAIN_HAND` is `1` and `SUB_HAND` is `2`, so the main hand is always processed first and the order is deterministic.

To reproduce: dual wield two weapons whose rank limit is above the current abyss rank and let the grace period expire.

Skipping items that left the slots during the same pass fixes it:

```java
for (Item item : getEquippedItems()) {
    if (!item.isEquipped()) {
        continue;
	}
```

PR #194, opened 2026-09-05
Fix missing cast time for item skills without casting_delay.

SVDNESS:
This fixes a regression introduced in PR #153.
Fixed item casting when `casting_delay` is not defined. The server now falls back to the associated skill's duration instead of casting the item instantly.
This restores the cast bar and animation for resurrection stones, teleport scrolls.

### DIFF
--- a/game-server/src/com/aionemu/gameserver/skillengine/model/Skill.java
+++ b/game-server/src/com/aionemu/gameserver/skillengine/model/Skill.java
@@ -379,8 +379,10 @@ public class Skill {
 	}
 
 	private int calculateCastDuration() {
-		if (getItemTemplate() != null)
-			return getItemTemplate().getCastingDelay();
+		if (itemTemplate != null) {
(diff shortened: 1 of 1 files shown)


Birthright50:
Any example?

SVDNESS:
> Any example?

Easy: https://aioncodex.com/ru/item/164000010/

Birthright50:
With this fix, you’ll break other items that shouldn’t have a cast, like 161000003, which had no cast.
The correct fix is to check the activation mode: Both -> use the item delay, Combat -> use the skill delay.

SVDNESS:
> With this fix, you’ll break other items that shouldn’t have a cast, like 161000003, which had no cast. The correct fix is to check the activation mode: Both -> use the item delay, Combat -> use the skill delay.

Yeah, sounds good. I’ll close my PR then. If you make any changes, you can include them in a new PR of your own.

PR #195, opened 2026-09-07
Recipes: fix opcode 158, complete S_ADD_RECIPE, and track recipe production counts.

SVDNESS:
## Background

Everything was verified against the unpacked NCSoft PTS 4.6 retail server (build 450319) and the 4.8 client.

### 1. Opcode 158 was mapped to the wrong packet

In your project opcode **158** was mapped to `CM_SHOW_BLOCKLIST`, which replied with `SM_BLOCK_LIST`.

In retail, opcode 158 is actually **`CM_RECIPE_LIST`**, a request for the player's learned recipes. The retail handler does not read any payload and simply returns the recipe list. The 4.8 client actively sends this packet with an empty body when the UI enters the in-game state.

The actual "show block list" request is **`C_QUERY_BLOCK` (168)**, and the 4.8 client never sends it. On 4.8 the block list is updated entirely through server push packets.

Opcode 158 has been remapped to `CM_RECIPE_LIST`, which responds with `SM_RECIPE_LIST`.

### 2. `SM_LEARN_RECIPE` (241) was missing one byte

Retail sends a 12-byte packet: header + `u32 recipeId` + `u8 count`

Your implementation only sent 11 bytes.

The client handler always reads both fields, meaning it was consuming an unrelated byte from the receive buffer as the count value.

The missing byte has been added.

`SM_RECIPE_DELETE` (242) and `CM_RECIPE_DELETE` (89) were also verified and already matched retail behavior.

### 3. Added support for recipe production counts

The missing byte is not a flag but the **remaining production count**:

* `0` = unlimited uses otherwise the value is decremented on each craft when it reaches zero, the recipe is removed.

The value comes from the `max_production_count` field in `combine_recipe.xml` and is persisted in the database.

Retail stores learned recipes as 5-byte records (`u32 id` + `u8 count`).

Previously, recipes were removed after a single craft whenever `max_production_count` was present, causing the 12 recipes with values greater than 1 to expire prematurely.

`RecipeList` now stores recipes as `Map<recipeId, remainingCount>`, and consumption follows retail behavior:

* one charge is consumed when crafting starts, not when it completes;
* combo stages do not consume additional charges;
* failed crafts do not refund the charge;
* when the count reaches zero, the recipe is removed, but the current crafting session is allowed to finish.

## Migration required

```sql
ALTER TABLE `player_recipes` ADD COLUMN `production_count` TINYINT UNSIGNED NOT NULL DEFAULT 0;
```


SVDNESS:
P.S. One more thing I forgot to mention: the following line should be removed from the `PlayerEnterWorldService` class:
Because the recipe list is now sent through the `CM_RECIPE_LIST` packet.

```diff
- client.sendPacket(new SM_RECIPE_LIST(player.getRecipeList().getRecipes()));

neon-dev:
@SVDNESS in case you're not aware, you can edit your PRs after they have been created. A PR is just a view of a branch. The changes you push to your PR branch will appear here as well :)

SVDNESS:
> @SVDNESS in case you're not aware, you can edit your PRs after they have been created. A PR is just a view of a branch. The changes you push to your PR branch will appear here as well :)

I’m aware, I sent the PR.
A day later, I remembered and was out of the house, so I wrote a comment so I wouldn’t forget.

PR #196, opened 2026-09-08
Fix effect ordering

Birthright50:
The list order decides which effect a dispel takes off first, and it follows the order casts ended, not the order effects land. An effect now reserves its place at the end of the cast and starts on the hit time as before, staying invisible to packets, dispels and conflict checks until then.

Also:
1) Remove dispellable buffs before rolling counter attack damage, with the removal broadcast synced to hit time.
2) Delay launched skills by 50 ms (on retail, the delay is also controlled by the `reserved3/reserved4` params, but they are 0 in all 4.8 skills, so this case was skipped. The minimum delay is always 50 ms)

Аfftected behavior:
Cast Body Root (3571) and instantly cast Root (1328): 3571 now stays ahead of Root in the effect list, even though it lands later.

Cast Magic Implosion (3555) -> Aegis Breaker (3740): the Aegis Breaker debuff appears earlier because the server reserves its effect slot before activating the `SkillLauncher`.

### DIFF
--- a/game-server/src/com/aionemu/gameserver/controllers/effect/EffectController.java
+++ b/game-server/src/com/aionemu/gameserver/controllers/effect/EffectController.java
@@ -93,11 +93,33 @@ public class EffectController {
 				chants.getFirst().endEffect();
 		}
 		put(mapToUpdate, nextEffect);
+		nextEffect.setSlotStarted();
 
 		nextEffect.startEffect();
 
 		if (!nextEffect.isPassive())
 			broadCastEffects(nextEffect);
+	}
+
+	/**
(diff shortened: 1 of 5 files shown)


neon-dev (commented):
Please check if my latest commit creates any conflicts I might have missed.
`EffectController.clearEffect()` has a slight behavior change, which looks safe to me.

I've also noticed a few methods in the EffectController where `isSlotReserved()` maybe should be checked as well, but I'm not too sure about them:
- `getAbnormalEffect()`
- `unsetAbnormal()`
- `resetDesignatedDispelEffect()`

PR #200, opened 2026-09-13
Reworked decomposable items

Birthright50:
### Data

1. Structure (branches, race/class/level restrictions, rewards) was imported from the 4.9 client (`client_disassembly_item.xml`, `client_disassembly_item_SetList.xml`).
2. Branch chances, weights and counts come from the 4.6 and 5.8 retail server data, matched by cName.

### Limitations

1. 281 entries use the 4.6 sets. 4.9 removed non-normal grade materials and old potion/scroll ranks from bundles (`AION_4_9v_Patch Notes_EN_24022016.pdf`, `AION_Patch_Notes_111815.pdf`), so where that is the only difference, 4.8 still had the 4.6 contents.
2. The other 127 entries that differ between 4.6 and 4.9 stay on 4.9. They also contain 4.8 changes (Illusion Godstones, Stigma Shards removed: `AION_4_8v_Patch Notes_EN_30092015.pdf`, `AION_Patch_Notes_061715.pdf`) or changes no patch note dates, and no source has the exact 4.8 set.
3. 58 rewards in 18 entries were dropped because those items don't exist in 4.8 client.
4. 36 entries are empty: 35 have no retail data in any version, and every reward of `188053631` is missing in  4.8 client.

[Patch notes ](https://github.com/beyond-aion/aion-patch-notes)


SVDNESS:
I recently worked on the same logic myself, using the **4.6 and 5.8 PTS servers** as references.

The implementation is generally very close to retail, but I found in your implementation two issues that are worth mentioning.

**1. The branch sort is inverted.**

`DecomposableItemInfo.afterUnmarshal` sorts by chance descending. Retail sorts **ascending** — the rarest branch gets the first roll, and the likely one acts as the fallback.

`CDisassemItemInfo::SortSetElements` is called at load time, and its comparator is:

```asm
movzx edx, word ptr [rdx+0x10e]   ; b.rate_
cmp   word ptr [rcx+0x10e], dx    ; a.rate_ < b.rate_
setb  al
```

For `only_one` entries the direction matters a lot: a 1% branch placed after a 90% branch is only reached in 10% of the rolls, so its effective chance becomes 0.1% instead of 1%. Dropping `.reversed()` restores the retail distribution.

**2. `GRANTED` / `NOT_GRANTED` are swapped.**

The last byte of the select response is the return value of `User::GiveSelectedDisassemblyItem`, and retail uses:

- `1` — the item was granted (set immediately before handing it out)
- `0` — not granted (the default, every early exit)
- `2` — refused by the drop limit (`lea ebx,[rax+2]` after `DropItemControlDB::AddCurCount` returns 0)

Same in both versions, so 4.x and 5.x agree: 4.6 success `0x343071` / limit `0x342c8c`, 5.8 success `0x3b9fe6` / limit `0x3b9c31`.

So `GRANTED` should be `1` and `NOT_GRANTED` `0`, and there is a third code for the drop-limit refusal if you want to surface it.

